PR #45, opened 2026-07-02
Fix modbus backoff jitter clock source, add retry tests

syphernl:
The retry backoff jitter read `asyncio.get_event_loop().time()`, a deprecated loop-internals access pattern, in both the read and write loops. Extracted a single `_backoff_jitter()` helper seeded from `time.monotonic()` and called from both sites. Same `+/-5%-of-delay` formula, no behaviour change.

Also replaced the bare `except Exception: pass` around the pre-reconnect client close with a debug log, so a close failure is no longer completely invisible.

First coverage for two paths that had none:

- `test_modbus_client_retry.py`: `_read_raw` and `async_write_register` retry/backoff, timeout and error-result retry, short-read rejection, invalid-arg guards, shutdown fast-exit, and that a connection error does not self-reconnect from inside the loop (coordinator owns reconnection, issue #361).
- `test_alarm_notifier.py`: notify on new bits, dismiss on all-clear, no re-notify on unchanged bits, fault-vs-warning severity.

Full suite: 556 passed / 10 skipped.

### DIFF
--- a/custom_components/omnibattery/infra/modbus_client.py
+++ b/custom_components/omnibattery/infra/modbus_client.py
@@ -8,6 +8,7 @@ from pymodbus.client import AsyncModbusTcpClient, AsyncModbusSerialClient
 from pymodbus.exceptions import ConnectionException, ModbusIOException
 import asyncio
 import inspect
+import time
 from typing import Optional
 
 import logging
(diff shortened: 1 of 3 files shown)


ffunes:
Thanks a lot for your contribution!

PR #53, opened 2026-07-03
Temperature-based charge power limiting

syphernl:
Optional feature (off by default) that lowers charge power when a battery gets hot and restores it as it cools.

Full power at or below the temperature limit, then a linear ramp down to a floor (% of normal charge power) across a band you set. Example: 40°C limit, 10°C band, 40% floor gives full power up to 40°C, 70% at 45°C, 40% from 50°C up. Smooth ramp, no hysteresis to tune. Per-battery on its own temperature. Missing sensor reading never blocks charging.

A hard floor stops the derate from commanding an unreliable low power: it clamps to each battery's minimum operating power from the driver (Marstek v2/v3 = 800 W, vA/vD/Zendure = 0), so v2/v3 floor at 800 W and the rest can derate to a stop.

Opt-in toggle applies the same curve to discharge, mainly to stay under the BMS hard over-temp cutoff.

Controls appear as normal entities (enable switch, discharge toggle, three sliders) plus a panel block, in all six languages.

Tested: 25 unit tests (incl. the real _battery_power_limit hook), full suite green (592 passed).

### DIFF
--- a/custom_components/omnibattery/__init__.py
+++ b/custom_components/omnibattery/__init__.py
@@ -39,6 +39,16 @@ from .const import (
     CONF_ENABLE_CHARGE_DELAY,
     CONF_DELAY_SAFETY_MARGIN_MIN,
     DEFAULT_DELAY_SAFETY_MARGIN_MIN,
+    CONF_ENABLE_TEMP_CHARGE_LIMIT,
+    DEFAULT_ENABLE_TEMP_CHARGE_LIMIT,
+    CONF_TEMP_CHARGE_LIMIT_C,
+    DEFAULT_TEMP_CHARGE_LIMIT_C,
+    CONF_TEMP_CHARGE_LIMIT_BAND_C,
+    DEFAULT_TEMP_CHARGE_LIMIT_BAND_C,
+    CONF_TEMP_CHARGE_LIMIT_FLOOR_PCT,
+    DEFAULT_TEMP_CHARGE_LIMIT_FLOOR_PCT,
+    CONF_TEMP_LIMIT_APPLY_DISCHARGE,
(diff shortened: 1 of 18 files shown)


ffunes:
Thanks a lot for this, @syphernl — really clean work. 🙏

I went through it in detail and the design is solid: it mirrors the MaxSocChargeManager cap-and-return contract, fail-safe passthrough when the temperature is missing/non-numeric, the hard min-power floor is handled correctly (min(min_power, limit) never raises a limit the SOC taper already pushed below it), off by default, per-battery, fully translated and well tested.

I pushed one small follow-up commit directly to the branch (fea6998): get_status() was implemented and unit-tested but never actually consumed anywhere, so the per-battery thermal-derate diagnostics (temperature, derate_factor, derating flag) were dead. I wired it into the system status sensor under a temperature_charge_limit attribute (gated on the feature being enabled), following the existing hourly_balance status pattern — so now the diagnostics are live and the test covers real code.

Full test suite stays green. 
Great feature to have — thanks again!

ffunes:
Hi @syphernl — heads-up: this PR broke integration startup, regardless of whether the feature was enabled.

ChargeDischargeController.__init__ computes the system capacity early on, which goes through _battery_power_limit → apply_temperature_limit → reads self._controller.temp_charge_limit_enabled. But the PR loaded the temp_charge_limit_* settings ~200 lines further down in __init__, so setup died immediately:

`AttributeError: 'ChargeDischargeController' object has no attribute 'temp_charge_limit_enabled'`

Note it crashed even with the feature off — the if not ...enabled check is itself the failing attribute access. Fixed in db6ebcc by moving the settings load above the capacity calculation.

The unit tests didn't catch it because they exercise TemperatureChargeLimitManager against a stub controller, never the real __init__ path.

PR #115, opened 2026-07-19
feat(dynamic-pricing): optional minimum arbitrage margin gate

syphernl:
Optional **Minimum Arbitrage Margin** for dynamic-pricing slot selection, plus a configurable **Round-Trip Efficiency**. Margin is empty by default; with it empty or 0, nothing changes.

## Why

A fixed charge ceiling says whether a price is low, not whether it is low enough to be worth a cycle. In winter the curve flattens: cheap hours still sit under a 0.25 ceiling, but there is no spread behind them, so the battery cycles for less than the round-trip losses cost.

Backtest on 590 days of Dutch EPEX (all-in), `charge if cheap_N <= peak_N * eff - margin` at margin 0.02, eff 0.866, versus a fixed 0.25 ceiling:

| N (charge hours) | fixed 0.25: margin / losing days | spread gate: margin / losing days |
|---|---|---|
| 2 | 105.9 / 5 | 120.5 / 0 |
| 4 | 177.4 / 11 | 200.3 / 0 |
| 6 | 219.5 / 19 | 246.3 / 0 |
| 8 | 225.9 / 25 | 260.0 / 0 |

Better at every window length, no losing days. Winter margin roughly doubles on 29% fewer charge days.

## How

`effective_charge_ceiling()` values the priciest remaining hours (as many as `hours_needed`, already computed by the engine), applies efficiency, subtracts the margin, and returns the stricter of that and `max_price_threshold` along with the arbitrage value for reporting. The engine calls it once and hands the result to `select_cheapest_hours()`, so the ceiling shown in the notification is by construction the one applied.

Two deliberate choices worth a look in review:

- The discharge estimate runs *before* the static ceiling filter. Those hours are by definition above the ceiling, so filtering first would leave nothing to price against and the gate would never open. Regression test pins it.
- The gate runs on the 00:05 evaluation only, not the evening recharge. That path is a deficit-driven safety top-up after a bad solar day, and by then the horizon holds no expensive hours, so the estimate collapses toward the charge price and the gate would refuse every recharge it exists to perform.

Margin 0 counts as disabled, since a NumberEntity cannot be cleared back to unset from the UI.

`CHARGE_EFFICIENCY` is untouched; it derates charge power to size a window rather than valuing stored energy. New setting defaults to 0.85, same value.

## Notes

- 730 passed, 10 skipped. 22 new tests in `tests/test_arbitrage_margin.py`, 3 more in `tests/test_pricing_notifications.py`.
- Two `number` entities, panel rows, tooltips in 6 languages, docs updated. The notification names whichever ceiling actually bound, so a skipped night points at the setting that caused it.
- Efficiency must be the *marginal* ratio, not gross. On my Venus E: slope 0.866 (r=0.985, 30d) vs gross 0.814. The gap is standby drain, which you pay regardless, so including it would refuse profitable charges. Documented.
- Possible follow-up: measure efficiency from the existing energy sensors and expose it as a diagnostic, so users do not have to guess the number. Left out here because auto-applying it on a fresh install is a different risk profile.


ffunes:
Thanks a lot for your contribution :)

PR #120, opened 2026-07-20
fix: PD deadlock at min SoC on grid sensors slower than the stale window

### LINKED ISSUE #117: PD deadlock: batteries never recharge at min SoC when the grid sensor is slower than the stale threshold
Thanks for Omnibattery — it has been driving 2× Marstek Venus E reliably. Reporting a deadlock I traced through the control loop, with a workaround that fixed it on my side.

**Version:** v1.0.0
**Setup:** 2× Marstek Venus E 3.0 (Modbus TCP, v3), main sensor = Enphase Envoy via the HA `enphase_envoy` integration

## Summary

When the grid sensor's update period (60 s) exceeds the controller's stale threshold (30 s), the PD loop can reach a state where it never issues a charge command again once discharge is blocked by min SoC — even with a large, sustained solar surplus.

Observed: **19 hours with 0.00 kWh charged** while exporting, and **3.81 kWh imported from the grid**, as counted by Omnibattery's own `sensor.omnibattery_daily_grid_at_min_soc_energy`.

## Root cause — three interacting behaviours

**1. `last_output_sign` is never cleared on the `No available batteries` path.**
When SoC hits min, that branch zeroes the command and `return`s *before* the end-of-cycle state update, so the controller stays latched on "discharge" indefinitely.

**2. Stale mode freezes the command at zero.**
`_stale_cycles > _max_stale_cycles` (15 × 2 s) sets `p_scale = 0`, so `new_power = previous_power` = 0. Note that `_stale_cycles` counts cycles *without a value change*, so a sensor legitimately reporting a steady value is indistinguishable from a dead one.

**3. A 0 W request is classified as discharge.**
`is_charging = new_power > 0` means zero falls to the discharge branch, so `_get_available_batteries(is_charging=False)` is queried — empty at min SoC — and the cycle early-returns. Those 0 W requests also reset `_zero_cross_since`, so the discharge→charge flip can never accumulate its 5 s hold.

## Evidence

One suppression per Envoy refresh, always at `0.0s`:

```
PD zero-cross hold: suppressing discharge->charge flip (912W, error=-904W) while actuator settles (0.0s/5.0s)
PD zero-cross hold: suppressing discharge->charge flip (892W, error=-884W) while actuator settles (0.0s/5.0s)
PD zero-cross hold: suppressing discharge->charge flip (973W, error=-964W) while actuator settles (0.0s/5.0s)
```

```
ChargeDischargeController: Sensor stale for 26 cycles (~52s). Safety recalculation.
Marstek Venus 1: Skipping discharge - blocked by min_soc
Marstek Venus 2: Skipping discharge - blocked by min_soc
ChargeDischargeController: No available batteries, setting all to 0.
```

## Reproduction

Grid sensor period > stale threshold, SoC at min, sustained surplus.

The deadlock is **conditional, not permanent**: it forms when the surplus is *steady* and self-resolves on any sharp load change (I saw it clear on a +1.03 kW import → −332 W export transition, which produced several minutes of continuously-changing sensor values). That is likely why it can go unnoticed — it silently repairs itself as soon as a big load switches on.

## Suggested fixes

1. Derive the availability direction from the **error sign**, not from `new_power > 0` — a 0 W command is not a discharge intent.
2. Reset `last_output_sign` on the `No available batteries` path, or move that path after the end-of-cycle state update.
3. Don't let stale-mode 0 W requests reset `_zero_cross_since` — hold the timer instead.
4. Warn at setup when the configured main sensor's update period exceeds the stale threshold. This is a silent misconfiguration today, and HA's `enphase_envoy` is hard-capped at `SCAN_INTERVAL = 60 s` — so every Envoy user is exposed to it.

## Workaround

A REST sensor polling `/ivp/meters/readings` every 5 s (light endpoint, ~64 ms measured; `/production.json` is unsuitable — ~2.5 s and rate-sensitive), used as the main sensor.

Result: `rms_error_w` **1771 → 43**, `pd_control_quality` `battery_limited` → `stable`, and `Sensor stale` disappeared entirely.

Happy to test a patch or provide more logs.

syphernl:
Fixes #117.

## Problem

At min SoC, a sustained solar surplus could leave the PD controller unable to
issue a charge command when the grid sensor updated more slowly than the stale
watchdog. The first fresh sample armed the discharge-to-charge settle timer, but
the stale 0 W safety recalculation between samples cleared that timer. Every new
sample therefore restarted the hold at 0.0 seconds.

The reported setup used Home Assistant's `enphase_envoy` sensor at a 60-second
cadence. That cadence is also unsuitable for reliable real-time grid regulation,
but the configuration previously accepted it without a visible warning.

## Changes

- Preserve an armed zero-cross timer when a stale safety recalculation freezes
  the previous command at 0 W. Fresh 0 W decisions retain their existing behavior.
- Keep `last_output_sign` latched on the no-available-batteries path. The 0 W write
  may not have reached an unavailable battery, so a later direction flip must
  still prove itself against the actuator settle window.
- Define sensor compatibility independently from the 30-second stale watchdog:
  grid-sensor intervals of 10 seconds or more are unsupported; 1-2 seconds is
  recommended.
- Measure cadence only from real sensor updates. Watchdog ticks with `elapsed=0`
  no longer reset the slow-sensor debounce.
- After three consecutive unsupported intervals, create a translated Home
  Assistant Repairs issue identifying the sensor and observed cadence. Clear the
  issue after three consecutive supported intervals.
- Add the cadence requirement to initial setup, reconfiguration, English/Spanish
  documentation, the changelog, and all maintained UI translations.

The PR targets `release/v1.0.1-beta`.

## Testing

- Full suite: `718 passed, 10 skipped`.
- Focused PD/cadence suite: `38 passed`.
- New coverage includes timer retention, real-zero non-regression, the reported
  slow-sensor flip sequence, watchdog ticks between 60-second samples, outage-gap
  debounce, the independent 10-second compatibility threshold, translated Repairs
  issue creation, and automatic issue clearing.
- JSON parsing, Python compilation, and `git diff --check` pass.
- Ruff reports the same 24 pre-existing findings in `__init__.py`; no new findings
  are introduced by this PR.


ffunes:
Thank you for the PR and for the careful root-cause analysis. The distinction between the intentional direction latch and the stale recalculation clearing the zero-cross timer was especially helpful, and the focused regression tests made the fix straightforward to validate.

I have built on your work by retargeting the PR to `release/v1.0.1-beta`, fixing the cadence debounce so watchdog ticks cannot reset it, and turning the slow-sensor warning into a translated Home Assistant Repairs issue with automatic recovery. The full local suite and all GitHub checks are green.

Thanks again for investigating the field behavior in detail and contributing a solid fix.

PR #137, opened 2026-07-22
feat(charge-delay): hold cushion-only shortfall for the cheapest hour

syphernl:
The energy-balance unlock fires at `net_solar < energy_needed × 1.3`. That 30% is a cushion, not the target, but the branch unlocks on the spot and never looks at prices, unlike the feasible-window path right below it. On a marginal-looking summer morning the whole PV self-charge then lands in the export peak while the midday trough is still ahead.

2026-07-22 here, 10 kWh forecast, released 07:00 with `unlock_reason: energy_balance`: needed 3.12 kWh, net_solar 3.42, factored edge 4.06. The forecast covered the target on its own, only the cushion was short. Battery filled 34% to 95% between 07:06 and 11:58 while export was worth €0.20/kWh; the 12:00-14:00 slots sat at €0.025-0.04. After 12:00 it was full, so the rest of the day exported at the bottom. About €0.35, no grid import involved (0.07 kWh all day).

**Change:** when `net_solar` is still at or above the bare `energy_needed`, hold for the cheapest feasible hour instead of unlocking. The window ends at the projected bare-balance crossing (new `safety_factor` arg on `_estimate_energy_balance_unlock_h`, default unchanged), capped by the time-backup edge, so the SOC target stays reachable. Both hold sites now write their status through one `_price_hold_release_h` helper. Real deficits (`net_solar < energy_needed`), days without usable price data, and non-price-driven predictive modes unlock immediately as before.

The projection is the cosine model, so it can disagree with measured production. Each cycle re-checks the measured net, so a bad projection unlocks itself.

6 cases added in `tests/test_charge_delay.py`; suite 711 passed, 10 skipped.

### DIFF
--- a/.github/ISSUE_TEMPLATE/battery_support_request.yml
+++ b/.github/ISSUE_TEMPLATE/battery_support_request.yml
@@ -1,0 +1,158 @@
+name: "🔋 New battery support"
+description: "Request support for a new battery brand or model"
+title: "[BATTERY SUPPORT] "
+labels:
+  - enhancement
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ## New battery support request
+
+        Thank you for helping us assess a new battery integration. You do not need technical knowledge: fill in what you know and select **I don't know** where appropriate.
+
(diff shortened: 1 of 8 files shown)


ffunes:
Thanks for your contribution!

PR #140, opened 2026-07-22
fix-issue-26-non-responsive

engelsofta:
Fixes a recovery loop in the non-responsive battery handling related to #26.

When a battery acknowledged a discharge command but remained in Standby, the recovery path created a fresh connection and sent an internal Standby command. On the following discharge cycle, this was treated as a new discharge episode and cleared the failure counter and wake state.

As a result, the affected battery could repeatedly reconnect without ever being marked as non-responsive.

Changes

Preserve the non-responsive episode during the recovery Standby command.

Ensure that charging power cannot be mistaken for delivered discharge power.

Monitor critical inverter and control telemetry separately from successful BMS reads.

Refresh all telemetry groups after reconnecting.

Report a failed RS485 re-enable as an unsuccessful wake attempt.

Keep non-responsive diagnostic attributes consistent after cooldown expiry.

Add regression tests for the recovery loop and wrong-direction power feedback.

Testing

773 tests passed

10 tests skipped

0 tests failed

Related to #26

### DIFF
--- a/custom_components/omnibattery/__init__.py
+++ b/custom_components/omnibattery/__init__.py
@@ -13,6 +13,7 @@ from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import Platform, CONF_HOST, CONF_NAME, CONF_PORT
 from homeassistant.core import CoreState, HomeAssistant, callback
 from homeassistant.exceptions import ConfigEntryNotReady
+from homeassistant.helpers import issue_registry as ir
 from homeassistant.helpers.device_registry import DeviceEntry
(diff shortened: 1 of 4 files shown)


ffunes:
Thanks for identifying the issue and for the detailed root-cause analysis. This PR mixed the #26 fix with files from a different branch and could not be loaded against its base branch, so I ported the focused fix onto the current release branch. The replacement preserves the intentional rollback of the issue #77 beta gateway workarounds: it does not restore `queued_gateway_compatibility`, while the tolerant power ACK from 1.0.0 remains unchanged. The clean replacement, including the 1.0.1b7 changelog, is #144. Superseded by #144.

PR #143, opened 2026-07-23
fix: preserve non-responsive recovery episodes (#26)

ffunes:
## Summary

Clean replacement for #140 and #142, rebased on `release/v1.0.1-beta` and limited to the issue #26 recovery path.

- preserve the active non-responsive episode across the recovery-only standby write, so the second failure threshold excludes instead of reconnecting forever
- require negative battery power before treating a discharge command as delivered
- track failures of safety-critical telemetry groups independently from successful BMS reads
- refresh every telemetry group after a fresh reconnect and report failed RS485 re-enables accurately
- expire non-responsive diagnostic attributes consistently after cooldown
- preserve the intentional rollback of the issue #77 beta gateway workarounds; this PR does not restore `queued_gateway_compatibility`, while the tolerant power ACK from 1.0.0 remains unchanged

The implementation also avoids two reconnects in the same poll when aggregate and critical-group failure thresholds are crossed together.

## Testing

- `python -m compileall -q custom_components tests`
- focused regression suite: 33 passed
- full pytest suite: 839 passed, 10 skipped

Supersedes #140 and #142.
Related to #26.

### DIFF
--- a/custom_components/omnibattery/__init__.py
+++ b/custom_components/omnibattery/__init__.py
@@ -3151,12 +3151,17 @@ class ChargeDischargeController:
         ignore_discharge_blockers: set[str] | None = None,
         bypass_blockers: bool = False,
         force_write: bool = False,
+        preserve_non_responsive_episode: bool = False,
     ) -> bool:
         """Set charge/discharge power for a single battery with ACK verification.
 
         ``force_write`` bypasses the bus-load skip-write so the command is always
         written, even when the battery's set-points already match — used by the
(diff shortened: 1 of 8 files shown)


ffunes:
The remote branch advanced to 99644ba with the 1.0.1b7 changelog, but GitHub did not synchronize this PR's cached head SHA. Replacing it with a fresh PR from the same branch so the changelog commit and CI are attached correctly.

ffunes:
Replacement PR: #144.

PR #195, opened 2026-07-28
fix: ignore transient Anker battery_power zero readings

wouterbouvy:
## Summary
- Filter one-shot near-zero `battery_power` (register 10008) samples on Anker while a non-trivial setpoint is still latched, so Home Assistant keeps the last good reading instead of flashing 0 W.
- Publish a real 0 W only after three consecutive near-zero polls (~6 s), so genuine idle/stop is still reported.
- Add a unit test covering transient reject vs sustained accept.

This addresses false HA dips that were observed while the Solarbank UI continued charging; failed Modbus reads already omit the key, but a decoded `0` was treated as valid telemetry.

## Test plan
- [ ] Unit: `pytest tests/test_anker_driver.py -q`
- [ ] On a live Anker Solarbank Max AC install, charge under automatic control and confirm HA `battery_power` no longer briefly flashes 0 W while the inverter keeps delivering
- [ ] Confirm a real stop/idle still reaches 0 W within a few poll cycles after the setpoint goes idle
- [ ] Confirm reconnect still clears the glitch filter state (fresh connection starts clean)

### DIFF
--- a/blueprints/central_status_webhook_reporter_blueprint.yaml
+++ b/blueprints/central_status_webhook_reporter_blueprint.yaml
@@ -1,0 +1,95 @@
+blueprint:
+  name: "Omnibattery - Central Status Webhook Reporter"
+  description: >
+    Periodically sends selected Omnibattery and Home Assistant entity states to
+    a central HTTP endpoint. The endpoint is configured once as a
+    rest_command; this blueprint configures the site identifier, the entities
+    to report, and the interval.
+
+    Before using this blueprint, define the selected rest_command in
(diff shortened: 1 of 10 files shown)


wouterbouvy:
See screenshot:
Scherm­afbeelding 2026-07-28 om 20 14 04

Before 20:10 contains this PR's fix. At 20:10 I reverted to current release where you can see the apparent issue (in blue). HA shows false 0-reads from time to time. The yellow line reflects the actual delivered power by the battery

PR #224, opened 2026-07-31
fix: guard states.get() against unset solar_forecast_sensor

brunosccosta:
_should_activate_grid_charging() called self.hass.states.get(self.solar_forecast_sensor) unconditionally. On installs with no solar panels (sensor left unset per the documented config, CONF_SOLAR_FORECAST_SENSOR defaults to None), this crashed with AttributeError: 'NoneType' object has no attribute 'lower' inside states.get() — surfaced to the user as a 500 on the "Re-evaluate Dynamic Pricing" button. Comment right above the call already noted "an unset forecast sensor must never reach hass.states.get()", but the guard was missing.

The conservative-mode branch a few lines below already handles a configured but dead sensor (forecast_state is None); this fix just stops an unset sensor from reaching states.get() at all, matching that existing branch's intent.

Added tests/test_grid_charging_no_solar_sensor.py: exercises _should_activate_grid_charging unbound with solar_forecast_sensor=None against a states.get() stub that reproduces the original crash if called with None, confirming the fix and guarding against regression.

### DIFF
--- a/custom_components/omnibattery/__init__.py
+++ b/custom_components/omnibattery/__init__.py
@@ -2687,7 +2687,11 @@ class ChargeDischargeController:
         # === STEP 4: Get Solar Forecast ===
         # Use the live sensor value directly — today's forecast updates throughout the day
         # and reflects improving accuracy as actual weather conditions develop.
-        forecast_state = self.hass.states.get(self.solar_forecast_sensor)
+        forecast_state = (
+            self.hass.states.get(self.solar_forecast_sensor)
+            if self.solar_forecast_sensor
+            else None
+        )
         if forecast_state is None or forecast_state.state in ("unknown", "unavailable"):
             # Conservative mode: assume zero solar, compare usable vs consumption
             total_available_kwh = usable_energy_kwh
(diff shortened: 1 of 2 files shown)


brunosccosta:
@ffunes I just started using this project! amazing integration. found one small bug, here's a fix. happy to collaborate more in the future, if i see anything else.

ffunes:
Thanks a lot for your contribution! Really much appreciate it!

brunosccosta:
great! i have one more small fix and another one, on a corner case with my marstek venus e 3.0 battery. Lemme explain a bit more my setup: my battery is connected via LAN cable and via HA i'm controlling using Omnibattery. I don't have solar but have dynamic contract. My goal is to offset consumption over the day and charge when it's cheaper. I want to cycle every day.

in case i send more MRs your way, you already know where i'm optimizing.

And for my future references, when do you merge to the main branch? which one should as starting point for next ones?

ffunes:
As long as existing functionality for users with solar panels remains intact, I'm totally fine with this! :)

For context, in the upcoming release, I'm introducing curtailment (for users with solar panels and dynamic tariffs) and grid charging optimization for users without solar panels.

Regarding my current workflow: I usually create a feature release branch, iterate on it with fixes and beta builds based on user feedback, and then merge it into main for the stable release.

Going forward, I’m happy to create branches after releasing 1.3.0:

- main: Current stable release. (1.3.0) nothing should be merged in to there unless it is a the step prior to publicly available release.

- patch branches (e.g., 1.3.1): Bug fixes for the current release.

- minor branches (e.g., 1.4.0): New features (or a mix of features and fixes).

Thanks for the input!

PR #228, opened 2026-08-02
fix: extend re-evaluate button's price horizon through end of tomorrow

brunosccosta:
## Summary

Simple fix for a corner-case situation: the "Re-evaluate Dynamic Pricing" button reuses `extended_horizon`, a flag originally written for a different case (HA restarting after the 00:05 evaluation window was missed, where a +12h floor guarantees a minimum planning window even for a restart minutes before midnight).

Pressed at an arbitrary time of day — e.g. 21:00 — `max(end_of_today, now+12h)` only reaches ~09:00 the next day, silently cutting off the rest of tomorrow's already-published EPEX prices, including cheap midday slots typical of solar-heavy markets (afternoon oversupply depresses the day-ahead price).

- Confirmed live: a 21:03 press picked slots at ~07:30–09:15 (right at the old boundary) while the actual cheapest slot of the day was at 14:00.
- Widening to end-of-tomorrow is always safe: `_parse_price_data` only returns slots that exist in the price sensor's data, so on days where the day-ahead publish is partial (observed in practice: some days only populated through noon the next day) this just yields fewer slots rather than erroring or waiting.

## Test plan

- [x] `tests/test_extended_horizon_reevaluate.py` added — pins the new horizon boundary against the old +12h floor, confirms the non-extended (00:05 daily run) path is unaffected, and exercises `_parse_price_data` unmocked to confirm both a previously-excluded cheap afternoon slot is now included and graceful degradation (no crash, no fabricated data) when tomorrow's publish is only partial

### DIFF
--- a/custom_components/omnibattery/pricing/engine.py
+++ b/custom_components/omnibattery/pricing/engine.py
@@ -607,9 +607,17 @@ class PricingManager:
         charging_needed = decision_data["should_charge"]
 
         # Step 2: Parse price data (always, even without deficit — for diagnostics)
+        # extended_horizon reaches through end of tomorrow rather than a fixed
+        # +12h: a +12h floor from an evening call (e.g. the manual re-evaluate
+        # button pressed at 21:00) only reached ~09:00 the next day, missing the
+        # rest of tomorrow's published prices entirely — including the cheap
+        # midday slots typical of solar-heavy markets. Requesting a horizon this
+        # wide is always safe: _parse_price_data only returns slots that actually
+        # exist in the price sensor's data, so on days where the day-ahead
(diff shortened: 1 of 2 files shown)


ffunes:
Thanks for investigating this and for adding the regression tests.

For now, I’m going to close this PR. The automatic Dynamic Pricing evaluation runs at 00:05 and considers the remaining slots for the current day, so tomorrow’s afternoon slots will be evaluated then.

Extending the shared `extended_horizon` could also affect startup recovery after a late HA restart, potentially selecting slots for tomorrow afternoon while leaving the current night insufficiently covered.

If immediate replanning from the manual button becomes a requirement, we can revisit it later with a separate horizon specifically for that button.

PR #252, opened 2026-08-08
feat: scale PD target grid power bounds with configured battery power

waelha:
## Summary

**PD Target Grid Power now scales with the installed battery power.** The slider is fixed at ±2500 W regardless of system size, so a three-battery Venus E system cannot be asked to import or export more than one battery's worth of power.

Its upper bound now becomes the sum of the configured per-battery maximum charge power, and its lower bound the negative sum of the configured maximum discharge power (positive = import from grid, the battery charges; negative = export to grid, the battery discharges). That gives ±7500 W for three 2500 W units, and asymmetric ranges for batteries whose charge and discharge ceilings differ (e.g. 2× Sessy 2200/1700 → -3400 … +4400 W).

When the optional system power limits are enabled, each direction is additionally narrowed to its configured cap — the cap can only narrow the per-battery sum, never widen it. So the advertised slider range and what the controller will actually allocate at runtime stay aligned; the entity never offers power the configuration does not permit.

The bounds follow every runtime change (the per-battery power sliders, the system-limit switch, the system caps) **without a reload**, because `native_min_value` / `native_max_value` are properties reading `config_entry.data` rather than frozen `_attr_` values.

Installations with no configured battery limits keep the current ±2500 W range, per direction.

## Why not reuse `MarstekController._effective_system_capacity`?

That one sums the *live* per-coordinator limits after temperature derate, SOC taper and slot ceilings for the current control cycle. A UI bound must advertise the *static configured* envelope — slider bounds must not jump around with battery temperature. The new helpers in `const/integration_const.py` therefore derive from `config_entry.data` alone (no hass, no coordinators).

## Also fixed along the way

- The help text in **every** locale claimed *"Range: -500 to +500 W"*, which matches neither the current ±2500 slider nor the new behaviour. Updated in `strings.json`, `translations/*.json` and the six `SYS_HELP` blocks in `marstek-panel.js`.
- The PD controller docs table had **no row at all** for this setting; added to `pd-controller.md` and `pd-controller.es.md`.
- The "Different grid target" and "Peak shaving recharge" blueprint selectors cap at ±2500 W, which would block multi-battery users from entering the new range; widened to ±15000 W to match the existing `system_max_*` ceiling.

## Testing

`tests/test_target_grid_power_bounds.py` (46 cases) covers: multi-battery sums, single-battery legacy parity, asymmetric brands, system caps narrowing each direction independently, caps unable to widen, legacy entries predating `CONF_ENABLE_SYSTEM_POWER_LIMITS`, per-direction fallback, step-grid flooring, the `min < max` invariant HA requires, and that every non-dynamic `CONFIG_NUMBER_DEFINITIONS` entry keeps its authored bounds.

Full suite on this branch: **1196 passed, 10 skipped**.

## Notes for the maintainer

- Branched from and targeting **`release/v1.3.0`**, not `main`.
- `CHANGELOG.md` and `manifest.json` are deliberately untouched — version numbering is yours. The summary above is ready to drop into a changelog entry if you want it.
- The **es / ca / de / fr / nl** strings are machine-assisted. The English is authoritative; please have a native speaker sanity-check the others before merging.


ffunes:
Thanks @waelha for the work on PR #252. The dynamic PD target grid power bounds and the related documentation improvements are being carried forward.

During integration, we also added a broader normalization of the battery power limits across all brands. To avoid maintaining two overlapping PRs, I’m closing this one as superseded by the follow-up PR.

Your original contribution is included and credited in the changelog. Thanks again!

Issue #251, opened 2026-08-08
Track battery IP changes via HA's DHCP discovery (opt-in, MAC-based) — design question before any PR


wilsto:
Hi @ffunes,

Same setup as [#245](https://github.com/ffunes/Omnibattery/issues/245) / [#246](https://github.com/ffunes/Omnibattery/pull/246) — 2 × Venus E v3 over Modbus TCP, Omnibattery 1.2.0.

**This is a design question, not a bug report, and I want your opinion before writing any code.**

I know you already answered this in [#123](https://github.com/ffunes/Omnibattery/issues/123):

> This is actually expected behavior rather than a bug. […] the integration relies on a fixed network address and cannot auto-discover or automatically track changes when the battery is assigned a new IP.

I'm not disputing that — it is a legitimate design choice, and the manual reconfigure flow works. I'm asking whether you'd be open to an **opt-in** mechanism on top of it, because we hit the failure mode in production and the usual mitigation did not prevent it.

## What we measured

On 2026-08-06 one of our Venus E v3 units started using a **second IPv4 address** while keeping the same MAC. The router's LAN browser reported both addresses as `l3connectivities` of a **single host**:

```
DC:04:5A:14:6B:33   192.168.1.181   (configured in Omnibattery)
DC:04:5A:14:6B:33   192.168.1.180   (the one actually in use)
```

Reconstructed from our DNS resolver's logs — the batteries resolve their cloud endpoint about once a minute, which makes an accurate retroactive timeline possible:

| When (UTC) | Address in use |
| --- | --- |
| until 2026-08-06 19:00 | `.181` — ~57 queries/h, continuous |
| 2026-08-06 19:00 → 2026-08-07 15:22 | `.180` — ~56 queries/h; `.181` drops to **zero** |
| from 2026-08-07 15:22 | back to `.181` |

So for roughly **20 hours the configured address was simply not in use by the device**. It came back on its own; nothing was reconfigured in Home Assistant.

**The part I think is worth your attention:** a **static DHCP reservation for that exact MAC was already configured on the router, and had been for weeks.** It did not prevent this. So "just set a DHCP reservation" — the natural answer, and the one we'd applied — was not sufficient here. I can't tell you *why* the device took a second address; I can only show that it did, with a reservation in place.

## Why it's hard to notice

[...]

PR #254, opened 2026-08-09
feat(config_flow): follow a battery's IP address by its MAC (opt-in)

wilsto:
Implements the design agreed in #251. Thanks for the detailed answer there — this follows it point by point.

Opt-in, per battery: when the router hands a battery a different IP address, Home Assistant re-finds it by its MAC and updates the connection instead of reporting it unreachable. Off by default; an install that never ticks the box behaves exactly as before.

## Against your list

| You asked for | Where it is |
| --- | --- |
| Opt-in per battery, no behaviour change when disabled | `_mac_tracking_schema()`; `tracking_enabled()` returns False without both the flag *and* a valid MAC |
| MAC detected automatically, with a validated manual fallback | `_mac_defaults()` reads Home Assistant's DHCP cache and pre-fills the field; `_validate_mac_tracking()` rejects anything that is not six hex pairs |
| Marstek TCP, Zendure Local, Anker TCP, Sessy — excluding serial, ESPHome, MQTT | `IP_BASED_BRANDS` plus `is_ip_based()`, which also excludes a *Marstek over serial* entry (brand alone is not enough) |
| Keep `{host}_{port}[_slave]`, reuse `_migrate_battery_registry_ids()` | `_async_apply_dhcp_lease()` calls it with the old and new endpoint; no identity scheme was changed |
| Only update on a unique MAC match with no endpoint conflict | `evaluate_lease()` returns `ambiguous_mac` for >1 match, `no_match` for 0, `endpoint_conflict` when the target endpoint belongs to another battery |
| Shared gateways, multiple IPs per MAC, no repeated reloads | Batteries behind one gateway share its MAC → `ambiguous_mac`, nothing moves. A renewed lease for the current address → `unchanged`, no reload. Two live addresses → see the design note below |
| Tests for preserved entity IDs/history and for disabled being a no-op | `test_migration_rewrites_unique_ids_but_never_entity_ids`, `test_disabled_tracking_leaves_the_entry_untouched` |

## Where the code went

The decision logic is a new pure module, `infra/mac_tracking.py` — no Home Assistant imports, so every guard is testable without an event loop. The config flow only calls it.

The two form fields land in **one** place per flow rather than in each brand's connection step: all four IP brands already converge on `async_step_battery_limits`, so the fields are added there and gated on `is_ip_based()`. That keeps the diff small and makes it impossible to miss a brand.

## Scope note — hardware I could not test on

I own two Marstek Venus E v3 and tested against them. **Zendure, Anker and Sessy are covered by unit tests only** — I have none of that hardware. The path is brand-agnostic (it rewrites `host` and reuses your existing migration), but you should know that before merging.

## A design choice I would like you to challenge

A device can hold two IPv4 addresses at once. I measured one of my Venus units answering on `.180` and `.181` **simultaneously for about 20 hours**, with a static DHCP reservation in place for that MAC — which is also why "just set a reservation" was not enough here.

So `evaluate_lease()` refuses to move a battery whose configured address **still answers** (`is_available` on the live coordinator): while the current endpoint works, a lease for a second address is extra information rather than a move, and switching would trade a working endpoint for an untested one and could flap between the two. A battery that has actually gone quiet is moved immediately.

If you would rather always follow the newest lease, dropping that guard is a two-line change and one test.

## Tests

43 new tests, `1211 → 1254` passing, 10 skipped, **no existing test modified**.

They run without the `hass` fixture on purpose: `pytest.ini` sets `-p no:homeassistant` and CI runs a bare `pytest`, so a fixture-based test would be skipped everywhere and prove nothing. The stubs are small and local to the test file.

## Two things worth flagging

**`battery_device_info` now reads `getattr(self, "mac", None)`.** Several existing tests build a stub coordinator with `types.SimpleNamespace` and read that property off it, so a plain `self.mac` breaks them. I preferred adapting the property over editing your tests.

**`async_step_dhcp` types its argument as `Any`** rather than importing `DhcpServiceInfo`, whose module path has moved across Home Assistant versions. It reads `.ip` and `.macaddress` defensively. Happy to pin it to the concrete type if you have a minimum HA version in mind.

## Translations

`strings.json`, `en.json` and `fr.json` carry the new strings. `ca`, `de`, `es` and `nl` fall back to English — I would rather leave those to a native speaker than guess.

The checkbox reads *"Re-find this battery by its MAC address if its IP address changes"*, with the description spelling out the difference between the two addresses, since the IP field sits right above it in the same form.

### DIFF
--- a/custom_components/omnibattery/__init__.py
+++ b/custom_components/omnibattery/__init__.py
@@ -195,6 +195,7 @@ from .const import (
 )
 from .control.charge_delay import ChargeDelayManager
 from .infra.coordinator import MarstekVenusDataUpdateCoordinator
+from .infra.mac_tracking import publishable_macs
 from .tracking.hourly_balance import HourlyBalanceManager
 from .tracking.non_responsive_tracker import NonResponsiveTracker
 from .control.weekly_full_charge import WeeklyFullChargeManager
@@ -7196,7 +7197,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     await async_get_backup(hass)
 
     coordinators = []
-    for battery_config in entry.data["batteries"]:
+    # A MAC shared by several batteries belongs to a Modbus gateway, not to a
+    # battery; publishing it would merge them into one registry device.
+    entry_macs = publishable_macs(entry.data["batteries"])
+    for battery_index, battery_config in enumerate(entry.data["batteries"]):
         coordinator = MarstekVenusDataUpdateCoordinator(
             hass,
             name=battery_config[CONF_NAME],
@@ -7230,7 +7234,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             battery_manual_mode_enabled=battery_config.get(
                 CONF_BATTERY_MANUAL_MODE_ENABLED, False
             ),
-            mac=battery_config.get("mac") if battery_config.get("track_mac") else None,
+            mac=entry_macs[battery_index],
         )
         # Physical phase is metadata for the safety limiter only.  It is never
         # used as an input to the global Grid 0 controller.
--- a/custom_components/omnibattery/__init__.py
+++ b/custom_components/omnibattery/__init__.py
@@ -7230,6 +7230,7 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             battery_manual_mode_enabled=battery_config.get(
                 CONF_BATTERY_MANUAL_MODE_ENABLED, False
             ),
+            mac=battery_config.get("mac") if battery_config.get("track_mac") else None,
         )
         # Physical phase is metadata for the safety limiter only.  It is never
         # used as an input to the global Grid 0 controller.
--- a/custom_components/omnibattery/config_flow.py
+++ b/custom_components/omnibattery/config_flow.py
@@ -37,6 +37,14 @@ from homeassistant.helpers.selector import (
     TextSelectorType,
 )
 
+from .infra.mac_tracking import (
+    CONF_MAC,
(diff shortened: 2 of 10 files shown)


ffunes:
Thanks for the thorough implementation. The overall approach looks good and all checks are green, but I found one issue that should be addressed before merging.
- Shared MACs can collide in Home Assistant’s device registry.
battery_device_info currently publishes the MAC in connections for every tracked battery. When multiple batteries share a Modbus gateway, they also share its MAC. Home Assistant treats connections as part of the device identity, so the second battery may be associated with the first device or cause a registry collision before AMBIGUOUS_MAC is ever evaluated.
Please detect duplicate normalized MACs within the configuration and avoid publishing CONNECTION_NETWORK_MAC for batteries whose MAC is ambiguous. This keeps the batteries as separate registry devices and safely disables DHCP tracking for that shared gateway. Please also add a regression test covering multiple batteries behind the same gateway.

Two additional non-blocking notes:
- _async_apply_dhcp_lease() evaluates each Omnibattery config entry separately. Multiple entries are unusual and may not be an officially supported setup, but the config flow currently allows them. We can either enforce a single Omnibattery entry explicitly or make DHCP matching global in a follow-up.
- dhcp.async_discovered_service_info() is only exported by recent Home Assistant versions (2026.8). Older supported versions silently fall back to manual MAC entry. Please consider documenting that limitation or adjusting the UI wording so automatic detection is not implied on every supported HA version.
Once the shared-gateway/device-registry case is fixed and the suite remains green, I think the PR should be ready for another review.

ffunes:
Thanks @wilsto for addressing the feedback so thoroughly and quickly. The shared-MAC case is now handled safely, the additional tests cover the expected scenarios, and all checks are green. Great work on the implementation and documentation — the PR looks ready to merge!

PR #305, opened 2026-08-19
Show EVSE power on the home animation

smos:
This change will show EVSE chargers on the dashboard.
Planned and changes with Gemini 3.7 flash

Used 1.4.0-b1 for the base.
Configured excluded "Peblar EVSE Power" entity is now shown as animatd line showing charging in teal. Screenshot attached to Pull request

Schermafbeelding 2026-08-19 141745
This would list any excluded device as a EVSE device, which may not be what everybody wants, but this does work for me. 
It's possible to split this out, or make a EVSE toggle per excluded entity to sum it on the dashboard.

Thoughts?

### DIFF
--- a/custom_components/omnibattery/__init__.py
+++ b/custom_components/omnibattery/__init__.py
@@ -258,7 +258,21 @@ def _excluded_devices_panel_config(data: dict, ent_reg) -> list[dict]:
         )
         devices.append(
             {
+                "name": device.get("name"),
                 "power_sensor": device.get("power_sensor"),
+                "activity_sensor": device.get("activity_sensor"),
+                "ev_charger_no_telemetry": device.get(
+                    "ev_charger_no_telemetry", False
+                ),
+                "allow_solar_surplus": device.get(
+                    "allow_solar_surplus", False
+                ),
+                "dynamic_power_control": device.get(
+                    "dynamic_power_control", False
+                ),
+                "cover_home_when_active": device.get(
+                    "cover_home_when_active", False
+                ),
                 "included_in_consumption": device.get(
                     "included_in_consumption", True
                 ),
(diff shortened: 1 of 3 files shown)


smos:
I now see that the description for the EVSE says "Zonne overschot" which is solar excess, but that isn't quite true, This is set when the excluded entity has the "Follows the sun toggle".

The Peblar does do this in a specific profile, but I had overriden it to "Normal".
Actually following the mode is not impossible, but that would require knowing about a lot of HA charger entities and their values.

It's probably fine for now, or just leave that out altogether.

ffunes:
Thanks for taking the time to work on this and for providing the example configuration.

We’re going to close this PR for now and won’t merge it in its current form.

We understand that it addresses the request from discussion #388 in Marstek Venus Energy Manager: showing an EV charger as a separate consumer in the energy-flow diagram. However, the current diagram has only one node and one line for all excluded devices.

Users may have several excluded devices, such as an EV charger, an AC unit, or other high-consumption appliances. Reusing the same node and automatically treating excluded devices as EVSE could either mislabel devices, combine unrelated loads under the EVSE label, or hide the other excluded devices. Adding a second line would require a broader redesign of the diagram layout and responsive behaviour.

The “solar surplus” label also does not necessarily represent the charger’s actual operating mode; it reflects a configuration setting, as you noted yourself.

The idea is valid, but it needs a more general design with explicit device types and enough space to represent multiple excluded loads. We may revisit it later as part of a wider dashboard redesign, but for now we are not going to merge changes of this kind.

Thanks again for the contribution and for helping explore the use case.

smos:
So how do you like to see this, keep the current excluded entities container and a seperate container for the EVSE items? Or the ability to toggle a single excluded item as being the EVSE charger(s)?

Considering that more then 1 EVSE isn't uncommon anymore, the seperate container would make sense. And if you mix units, not all of them support FollowTheSun.

The current Exclusion logic works for me as it is.

PR #315, opened 2026-08-20
Update peak_shaving_limit_sync_blueprint.yaml

Wimpie-ccc:
Updated blueprint with an offset and a minimum peak value.

The offset will prevent upside creep of the monthly peak value.

The minimum peak value is useful at the start of the month. And could allow the user to draw more power because the grid provider always bills minimum this value (as in Flanders: always minimum 2500).

### DIFF
--- a/blueprints/peak_shaving_limit_sync_blueprint.yaml
+++ b/blueprints/peak_shaving_limit_sync_blueprint.yaml
@@ -1,9 +1,9 @@
 blueprint:
-  name: "Omnibattery - Peak Shaving Limit Sync"
+  name: "Omnibattery - Peak Shaving Limit Sync (with Offset)"
   description: >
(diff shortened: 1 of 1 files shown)


ffunes:
Thanks for the update. The offset and minimum peak options look useful, but I found one issue before merging:

When the monthly peak sensor is `unknown`, `unavailable`, or uses an unsupported unit, the automation falls back to the minimum value and writes it to the limit. With the defaults, this can set the limit to `2300 W` during a temporary sensor failure.

Could you keep the automation from updating unless the sensor state is numeric and its unit is either `kW` or `W`? It would also be good to add tests for invalid sensor states and values below the entity's valid range (`500–20000 W`).

Once that is fixed, the PR should be ready to merge.

PR #327, opened 2026-08-23
Add Huawei SUN2000 + LUNA2000 driver

sphings79:
Adds support for Huawei SUN2000 hybrid inverters with LUNA2000 storage.

Everything here was developed against live hardware — a SUN2000-8K-MAP0 with
13.8 kWh of LUNA2000 behind an EMMA-A02 — and every figure in the driver is a
measurement rather than a datasheet value. The full write-up is in
`site-docs/reference/driver-assessment-huawei.md`, filled in against the
repository's own driver-requirements template. Verdict there: **suitable with
limitations**, and the limitations are named.

## What it adds

- `drivers/huawei.py` — the driver, and `infra/huawei_modbus_client.py`, a small
  FC03/FC16 client.
- Telemetry over native Modbus. Control takes either of two paths, chosen per
  battery: the `huawei_solar` integration's services (default), or the same
  four-register sequence written directly, which removes the dependency
  entirely. Both are verified on hardware.
- Setup discovers what it can: the Modbus unit id is found by scanning, and a
  cascade of inverters is resolved by asking the user which one the battery
  belongs to.
- 152 tests, with register fixtures captured from the reference unit, so the
  decoding assertions double as a regression test for the register map.

## What it changes outside the driver

Three touchpoints, and they are the parts worth your attention:

1. **`BatteryDriver.dynamic_discharge_limit_w(data)`** (`drivers/base.py`) — new,
   defaulting to `None`, so every existing driver is unaffected. A DC-coupled
   hybrid shares one inverter between battery and PV, so the discharge power
   actually available is whatever the AC rating has left after PV. At 7 kW of PV
   on an 8.8 kW inverter the battery can contribute 1.8 kW whatever its 7 kW BMS
   allows — and load sharing, allocating against nameplate limits, hands it 74 %
   of the deficit and starves the batteries that could have delivered.

2. **`config_flow.py`** — a connection step and a cascade-selection step for the
   brand. Self-contained, but duplicated across the config and options flows the
   way the existing brands are; a test asserts the two copies stay identical.

3. **`infra/coordinator.py`** — a brand branch and the manufacturer name, plus
   the device entry now carrying a serial where a driver reports one.

Panel: each populated battery pack appears under device information with its own
serial and firmware, alongside the inverter and the power module.

## Limitations, stated plainly

- **One installation, one firmware pair.** Everything is single-sample.
- **A zero set-point releases this battery rather than pinning it.** Pinning it
  at 0 W does hold it, and was the first implementation — but a battery that
  cannot absorb its own PV makes the inverter derate the strings instead
  (observed: 4757 W to 55 W, for over half an hour of daylight). Releasing hands
  the meter back to the inverter's own regulation, which is the cheaper
  side-effect but is a real one.
- **No cell voltages.** A LUNA2000 reports per-pack values only, so balance
  monitoring and the 100 % voltage taper are disabled for this brand.
- **Cascaded inverters are guarded against but untested** — the reference bus
  holds exactly one.
- **Direct Modbus writes are off by default.** Verified on one installation.

## Notes for review

- A Huawei inverter accepts a single Modbus connection. The setup says so and
  points at a Modbus proxy; the reference installation runs one, and this driver
  read continuously while `huawei_solar` held its own connection.
- The version in `manifest.json` is untouched — that is yours to set.

Happy to split this up, rework any of it, or take it in a different direction if
the shared-code changes are not what you would want.


ffunes:
Hi,

First, thanks a lot for this contribution, it is awesome. I reviewed PR #327. The Huawei-specific tests pass (152), and the full test suite passes (1443 passed, 10 skipped), but I found four issues that should be addressed before merging:

1. P1 — Huawei backup mode does not disable PD control.
Huawei publishes `backup_function` as `"Off-grid"`, `"Ready"`, or `"Disabled"` (`drivers/huawei.py:633-640`), while `_is_backup_function_active()` only recognizes numeric `0` (`__init__.py:2612-2628`). During an outage, Omnibattery may continue sending charge/discharge commands. The shutdown guard has the same mismatch.

2. P1 — Direct-write mode is not released during shutdown.
`HuaweiSolarDriver.standby()` always calls the `huawei_solar` service (`drivers/huawei.py:1061-1072`). In direct-write mode, the integration or device may not exist, so the call fails and the last forced command can remain active for up to 10 minutes. Please use the direct Modbus release sequence in this path and add a regression test.

3. P2 — The options flow uses Marstek limits for Huawei.
The Huawei branch is missing from `async_step_battery_limits()` in the options flow (`config_flow.py:4083-4104`). It falls back to the Marstek v2 limit of 2500 W, which can silently reduce an existing Huawei configuration when options are saved.

4. P2 — Huawei reconfiguration is routed through the Marstek flow.
`async_step_reconfigure_battery()` has no Huawei branch (`config_flow.py:2650-2680`), so an existing Huawei battery receives the Marstek form and connection probe instead of Huawei settings.

Please address these points and add regression tests covering the backup state, direct-write shutdown, options limits, and reconfiguration flow.

sphings79:
Thank you for the review — all four hold up, and I have pushed fixes with regression tests for each.

**1. Backup mode (P1).** Fixed. `backup_function` now goes through one helper that answers both shapes: the register convention where 0 means the switch is on, and the three states this driver names for a hybrid (`Off-grid`, `Ready`, `Disabled`). The shutdown guard shares it, so the two cannot drift apart again. A test pins every state and a second one asserts the driver publishes no fourth spelling that would silently disable the guard.

**2. Direct-write release (P1).** Fixed, and thank you for catching it — I hit the same bug on live hardware the day after opening this, from the other end. A forcible discharge written at 04:32 was still in force at 09:22, through a cloudless sunrise, because the release had gone out over a service path that has no device to address in direct-write mode and fails silently during shutdown. On this hybrid a latched command also keeps the MPPT tracker down, so the array produced nothing all morning: 374 V at 0.00 A, and 11.4 A twelve seconds after the registers were cleared by hand. `standby()` now releases over whichever path the set-points took.

That incident also produced a change worth flagging, since it narrows what this driver claims. A forcible command on a SUN2000 is not a request but a ceiling: the inverter produces exactly what it was told and curtails the rest of the roof — measured with a 315 W charge standing, 288 W harvested against 5054 W six seconds after it ended, while a separate array on the same roof held steady. The driver therefore commands nothing at all while the strings carry voltage, and leaves the inverter to its own regulation during daylight. **A Huawei battery is under this integration's control after dark only.** That is now in the assessment (§13.8) and in the decision report rather than buried.

**3. Options-flow limits (P2).** Fixed — the Huawei branch was in the setup flow only, so saving options would have cut a 7 kW battery to 2500 W.

**4. Reconfiguration (P2).** Fixed. There is now a `reconfigure_battery_huawei` step with its own form and translations, instead of the Marstek form with a battery version and a slave id meaning something else. It re-probes, keeps the registry ids on an address change, and re-detects an EMMA.

Full suite: 1464 passed, 10 skipped. Huawei-specific: 173.

One note for anyone reading this later: the EMMA is not a requirement. It is needed only for the fast grid-meter sensor (register 31657, live per request, where `huawei_solar` publishes the same figure on a 30 s coordinator). The driver itself works on any SUN2000 with a LUNA2000, and the daylight limitation above applies either way.

Separately: the installation this was developed on runs a Huawei alongside an AC battery, and that combination turned up three control-layer problems that are not Huawei-specific — a slow battery never finishing because the surplus is shared by power rating rather than by the room left, and a second regulator on the same meter leaving one battery idle. I have those working locally and will offer them as a follow-up on top of this one, so this PR stays a driver.

ffunes:
Sorry — I initially said that all requested points were covered, but after reviewing the updated head more carefully I found a few remaining issues.

The four original review comments are fixed, and the tests pass: 173 Huawei-specific tests and 1464 tests in the full suite.

Before merging, could you please address the following?

1. Update the branch with the latest `release/v1.4.0` and resolve the conflict in `custom_components/omnibattery/infra/coordinator.py`, then rerun the full suite.
2. In Huawei reconfiguration, validate that the selected `huawei_solar` device belongs to the inverter identified by the Modbus serial.
3. Migrate entity and device registry IDs when the Huawei `slave_id` changes.
4. Handle multiple cascaded inverters during reconfiguration instead of returning `cannot_connect`.
5. Clear a stale `huawei_emma_slave_id` when reconfiguring to an endpoint without an EMMA.
6. Update the outdated module documentation in `drivers/huawei.py` regarding direct writes and `apply_setpoint(0)`.

Sorry for missing these points in my first pass. Once these are addressed and the rebased branch passes the suite, I think the PR will be ready.

sphings79:
No apology needed — the second pass caught real things, and four of the six are guards the setup flow had grown one at a time while reconfiguration, written later, had none of them. All six are addressed.

**1. Rebase.** Merged `release/v1.4.0`. The conflict in `coordinator.py` was a single line — the brand list deciding which drivers skip the version-based register map — and both sides belonged in it (`zendure, anker, hoymiles, huawei` plus the `zendure_model` assignment).

**2. Device/serial cross-check on reconfiguration.** Added, the same check setup makes: the chosen `huawei_solar` device must belong to the inverter the Modbus address points at, or the step refuses with `huawei_device_mismatch`.

**3. Registry migration on a slave-id change.** Added. `_migrate_battery_registry_ids` already took the slave arguments; the Huawei step was passing only host and port, so changing the unit id renamed every entity and stranded its history.

**4. Cascades during reconfiguration.** The cascade step is now reachable from reconfiguration as well, with its own form and translations, instead of `cannot_connect`.

**5. Stale EMMA unit id.** Now re-detected on every reconfiguration and cleared when the new endpoint has none. Carried over it would have the driver reading a grid meter that is not there, or whatever else answers on that id.

**6. Module documentation.** Rewritten. It still described the first design — services only, and a held zero expressed as a forcible charge at 0 W — and both are now the opposite of what the driver does. The daylight limitation is stated there too; it belongs at the top of the file, not only in the assessment.

Suite after the merge: **1764 passed, 10 skipped, 2 failed**. The two failures are `test_remaining_fallback_is_conditioned_on_today_consumption` and `test_full_day_history_is_restored_without_invalidation`. Both fail on a clean checkout of `release/v1.4.0` with this branch nowhere in sight — I checked in a separate worktree before touching anything, so I have left them alone rather than adjust tests that are not mine. Say the word if you would like me to look at them in their own change.

Huawei-specific tests: 180.

ffunes:
Hi,

Last changes before merging. Would you mind fixing the errors in the checks, please?

Thanks

sphings79:
Both checks are green now.

**Hassfest.** My fault, and a good rule: a translation string may not contain a URL, and the Modbus-proxy hint carried one into every language file at once. The link is passed as a description placeholder now, so it still reaches the screen. A test walks every translation file for URLs, since this is the kind of thing that is otherwise found late and by someone else.

**pytest.** The two failures were stale test harnesses, not product faults, so only the tests changed — I did not touch behaviour that is not mine:

- `test_full_day_history_is_restored_without_invalidation` builds its tracker with `__new__`, which skips `__init__`. History loading filters vacation periods out of the restored days, so without `_vacation_periods` it failed on the missing attribute rather than on the logic it means to check. One line in the factory.
- `test_remaining_fallback_is_conditioned_on_today_consumption` stubs `forecast_energy_between` on the profile, while `_profile_remaining_consumption` asks the tracker for `forecast_consumption_between`. The forecast never arrived, so the fallback adjustment the test asserts was never reached — and `adjust_remaining_fallback_energy(10.0, 20.0, 15.0, 12.0)` returns `(7.0, -3.0)` when called directly, exactly what the test expects. The stub now sits on the tracker.

Happy to move those two into their own commit or PR if you would rather keep them separate from the driver — say the word and I will split it.

Full suite: **1768 passed, 10 skipped**. Huawei-specific: 182.

PR #335, opened 2026-08-25
Charge order, primary battery, and two control-layer fixes

sphings79:
Rebased onto the current `release/v1.4.0` (`4ed3b29`). #327 has merged, so the diff is the changes proper.

Three control-layer problems, none of them Huawei-specific. They surfaced on an installation running a hybrid inverter beside an AC battery, because that pairing breaks assumptions a fleet of like batteries never tests: the two differ by an order of magnitude in charge power, and one of them is regulated by something other than this integration.

## The slow battery never finished

A surplus was shared between batteries **in proportion to their power rating**. That gives the slow one the smaller share — while it is the one needing the most hours. Measured: a 2.5 kW battery offered 1053 W of a 4 kW surplus beside a 7 kW inverter taking 2947 W. The inverter was full by early afternoon; the other was still at 52 % at sunset with a kilowatt going to the grid.

Charging now shares by **remaining energy**, capped at each battery's power limit with the excess redistributed. Every battery then aims at the same finish time, which is the earliest any of them can be done, and nothing is exported while capacity is free. Discharge is untouched: there the question is who can deliver now, and the power rating answers it.

Which battery leads follows the day. With sun enough for all of them, the one with the longest time to full goes first, since it is the one at risk. On a day the forecast cannot fill them, the DC-coupled one leads instead: its photovoltaic input never leaves DC, so scarce kilowatt-hours lose the least there. The verdict is latched against a wandering forecast.

No new setting for daily consumption: the integration already keeps a seven-day average of measured house consumption, and a figure entered by hand would sit next to a measured one.

## A second regulator took everything

Feedback control can only act on an error that already exists. A hybrid inverter running its own self-consumption removes that error first, so this integration correctly commanded nothing — and the second battery never ran. The grid regulation looked excellent while half the storage sat idle: 5834 W of sun, the hybrid at 83 %, the other battery at 17 % and never asked for anything.

Optionally, a nominated **primary battery** is handed the real quantity directly rather than waiting for the meter to deviate: the load the batteries have to cover, or the surplus that wants storing. The other regulator then finds nothing to do and stays available as a fallback instead of being configured away. It is a floor, not a target — the loop may always ask for more, and every downstream blocker still has the last word.

Both off by default, both surfaced in the panel with their figures shown whether or not they are on, so they can be checked against a meter before being switched on.

## Discharging into a surplus

Not a feature, a rule that was missing. With two regulators on one meter, one battery charging and another discharging cancel out: the grid reads zero, the deadband holds the standing command, and the arrangement is stable. Measured: 1391 W of PV over a 529 W house, one battery taking in 1110 W while the other gave up 205 W, the meter at 3 W — 829 W of surplus making a round trip through two conversion losses.

A discharge is now refused whenever the load the batteries must cover is negative. Latched with a band on the way in and none on the way out, so a cloud edge cannot toggle the battery in step with the light, while a real deficit hands the house back at once.

## Discharging past the house's own load

The mirror of the feedforward, and missing for the same reason. The loop chases the raw meter to zero, so anything else drawing on that meter becomes this integration's job — including a second battery being charged by its own energy manager. Covering that discharges one battery into the other through two conversions, and where the other manager fills the gap from the grid, it is paid for twice. Measured at 21:39 with no sun: the EMMA drew 211 W to charge the hybrid while the AC battery discharged 912 W against a 787 W house — the difference chasing a load that was never the house's.

A discharge is now capped at the load the house has actually left uncovered: the same figure the feedforward uses as its floor, which already has every battery's own contribution removed. A deadband of slack keeps measurement noise from clamping the command on and off every cycle. Unreadable, and nothing is trimmed; a surplus belongs to the guard above, which vetoes rather than caps.

## Also

A device reporting a power ceiling of zero is no longer believed. A Marstek came back from a restart with both power registers at 0; adopting that shut it out of every allocation, and nothing would have written those registers again because the controller had stopped addressing a battery it believed could do nothing.

And a battery whose device counts its own daily energy no longer zeroes the system totals. Those totals require every battery's figure to be marked as belonging to today, and only the derived counter stamped that mark — so one native counter in a fleet produced 0.00 kWh charged and discharged on the overview, with 10.3 and 3.97 kWh sitting in the per-battery sensors.

And the primary battery no longer overrides the ordering it was meant to sit on top of. Left on automatic it was chosen by raw SOC, and that outranked the 5 % edge the discharge ladder gives a battery already running: two batteries within that of each other traded places from one cycle to the next, which is the swapping the edge exists to prevent. Both now read one shared sort key. A battery under manual ownership is never the primary either — the distribution leaves it out of the ladder, so a feedforward measured against its rating would have been handed to the batteries that were left.

## Testing

Each behaviour above has tests carrying the measured numbers, and each was checked to fail against the previous code.

Everything here ran on one installation. The forecast-driven half of the charge order in particular has had far less exposure than the rest, and I would rather say so than let it look settled.


ffunes:
Thanks for the work. The overall approach looks solid, and the test suite currently passes, but I don’t think this is ready to merge yet.
Before merging, please:
1. Rebase the branch onto the current release/v1.4.0 and resolve the merge conflicts. GitHub currently reports the PR as dirty.
2. Fix the solar forecast handling in _charge_outlook_kwh(). The configuration accepts both kWh and Wh, but the new code always interprets the raw sensor value as kWh. It also subtracts the full daily consumption instead of consistently using the remaining forecast and remaining consumption for the current time horizon.
3. Add tests covering:forecasts reported in Wh;
 - full-day versus remaining-day forecast values;
 - behaviour at different times of the day;
 - feed-forward enabled without an explicitly selected primary battery.

4. Clarify or enforce the expected behaviour when feed-forward is enabled but no primary battery is selected.
5. Re-run the complete test suite, and ideally validate the forecast-driven priority logic on a real installation before merging.

The current tests are encouraging, but the forecast unit/horizon issue could cause incorrect charge prioritisation in real installations.

sphings79:
Rebased and updated — all five points addressed.

**1. Rebase.** The branch is now two linear commits on current `release/v1.4.0` (c9eee90), no merge commits, conflicts resolved. GitHub reports it mergeable and all three checks pass.

**2. Forecast handling.** You were right, and the fix was to stop hand-rolling it: `_charge_outlook_kwh()` now delegates to `solar_forecast.read_remaining_solar_kwh()`, which already knows the units, the optional "remaining today" sensor, and the whole-day conversion. It treats a `fallback`/`unsafe_zero` answer as "nothing readable" rather than as 0 kWh of sun — taken at face value that would call every day scarce. Consumption is trimmed to the same horizon, and both halves now run off a single timestamp: the reader keeps its own clock, so left alone it could convert the solar side against a different minute than the consumption side was trimmed to.

**3. Tests.** Wh and unit-less sensors; whole-day versus remaining-day; four pinned times of day; feed-forward with no nomination. The outlook tests pin the clock — since the window-fraction conversion landed, an unpinned test follows whatever hour the suite runs at, which is how two of mine broke on your newer base.

**4. No primary selected.** It falls back to the battery the ordinary discharge ordering would have picked anyway — the fullest one — so enabling feed-forward changes *when* a battery is asked, not *which*. `None` only when no battery can serve at all. Covered by `test_without_a_nomination_the_ordinary_choice_is_used`.

**5. Suite and real installation.** 1916 pass locally, CI green. Validation ran on the reference installation (Marstek Venus D + SUN2000/LUNA2000 behind an EMMA) and is written up in `site-docs/reference/primary-battery-field-evidence.md`.

That validation earned its keep: it exposed a further defect, now fixed here. Scarcity was measured against the room left in the *whole fleet*, which asks "is there enough to fill everything" — on a mixed installation almost never true. Every day came out scarce, so the AC battery was passed over on all of them, and its own empty capacity was most of what made the day look scarce. It sat at its 12 % floor taking 0.02 kWh in a day. Measured against the DC battery's room instead, the next day it took 13.99 kWh.

The honest limits are in the document: two days, one brand pairing, the sunnier day is the one after the change, no full charge was reached, and the handover between two batteries that both had room has not been observed yet.

sphings79:
The failing `Tests / pytest` check comes from the base branch, not from this PR.

Every test errors at setup with the same message:

```
pytest_homeassistant_custom_component/plugins.py:342: in enable_event_loop_debug
    asyncio.get_event_loop().set_debug(True)
RuntimeError: There is no current event loop in thread 'MainThread'.
```

1989 errors, 31 passed. The chain:

- `pytest.ini` carries `addopts = -p no:homeassistant`, with the comment explaining that the plugin's autouse fixtures make the whole suite error at setup.
- ad0b54e changed the CI step to `pytest -o addopts=""`, which drops that flag and loads the plugin for every test.
- Its autouse `enable_event_loop_debug` fixture calls `asyncio.get_event_loop()`. Under `pytest-asyncio` 1.3.0 with Home Assistant's `HassEventLoopPolicy` there is no loop to get, so it raises before any test body runs.

This PR touches none of it — the diff against the base is empty for `pytest.ini`, `tests/conftest.py`, `requirements-test.txt` and `.github/`. It was green on 2026-08-28 with the same tests, and the last green run on any branch (2026-08-30 17:26) installed the identical versions — pytest 9.0.0, pytest-asyncio 1.3.0, homeassistant 2026.2.3 — but predated ad0b54e, so it still ran with the plugin disabled and finished in 8 seconds.

Any pull request based on the current `release/v1.4.0` should fail the same way, so this is worth fixing in the base rather than here. Happy to open a separate PR for it if that helps — the two options I see are creating the loop the plugin expects in a session fixture, or keeping `-p no:homeassistant` for the unit suite and running the flow tests as their own step.

ffunes:
The evidence here is worth more than the diff, so let me be direct about what I want to do with it.

#327 was safe to merge because a driver only affects whoever configures it. This one changes the control loop for every installation, and two things in it are why I want to own that part rather than iterate on it in review:

- `_apply_surplus_guard` and `_apply_discharge_ceiling` (`__init__.py:9101-9108`) run unconditionally for everyone and compare `_uncovered_load_w` against 0 instead of `compute_active_target()`. That cancels `curtailment_predischarge` (`pricing/engine.py:2181`), which sets a negative override on purpose to build headroom before a negative-price window — vetoed outright under sun, trimmed to house load at night. Same for any negative `pd_target_grid_power` (the number accepts −2500 W).
- `_uncovered_load_w` returns `None` only when *no* battery is readable. With one battery off comms but still delivering (the hardware holds its last command), its `ac_power` drops out of the sum, the uncovered load understates the real one, and the ceiling trims the healthy battery until the house imports.

So, proposal — three pieces instead of one:

1. **The two coordinator fixes in their own PR, and I merge them this week.** The zero power-ceiling guard and the native daily-counter stamping are ~120 lines, self-contained, and clearly right. No changes needed from my side.
2. **The Huawei release in its own PR.** `apply_setpoint` dropping to 0 whenever `_pv_lit` is a driver-wide behaviour change that isn't in the description, and >100 V on a string is a light test with no look at harvested power — an overcast dawn at 300 V and 50 W blocks discharge too, when the house needs it. Gate it on harvested power and it's yours to land: it only affects Huawei users.
3. **The control-layer work as a spec, which I implement.** Charge order, remaining-energy allocation, primary battery, feedforward, surplus guard, discharge ceiling. Your PR description is already most of that spec — keep the measured numbers, they're the part I can't reproduce. I'll build it in `control/` so it respects the active target and the existing overrides.

What I'd genuinely rather have from you is the mixed installation. Nobody else here has a 7 kW hybrid next to an AC battery with an EMMA on the same meter, and readings like "21:39, EMMA pulling 211 W while the Marstek discharged 912 W against a 787 W house" are what make these bugs findable at all. Testing that pairing is worth more to this project than the loop code.

I'll leave this open until the two split PRs are up, then close it pointing at them.

sphings79:
Agreed on the split. #380 is up with the two coordinator fixes; the Huawei one follows.

Both of your objections are right, and I checked them against the code rather than taking them on trust. Since you're implementing this part, here is what I know that isn't in the diff — including a third defect I found while verifying yours.

### 1. The guards ignore the active target — confirmed

`_apply_surplus_guard` and `_apply_discharge_ceiling` compare against 0, and every other branch in that control cycle compares against `compute_active_target()`. The deadband shortcut twenty lines earlier is centred on `active_target`; my two guards were the exception, and I didn't notice because the reference installation never sets a negative target.

The consequence is exactly as you describe. `curtailment_predischarge` sets a negative override on purpose (`pricing/engine.py:2182`, priority 5). Under sun the surplus guard zeroes it outright, so the headroom is never built and the negative-price window arrives with a full battery. At night the ceiling trims it to house load — on the reference installation that turns a 2500 W predischarge into 787 W.

The fix I'd have made: both guards take `compute_active_target()` as the floor instead of 0, so they can only ever trim toward a target someone deliberately set, never past it.

### 2. A silent battery understates the uncovered load — confirmed

`_uncovered_load_w` (`__init__.py:865`) `continue`s past an unreadable coordinator and sets `seen` from whichever ones answered, so it returns a number as soon as *one* battery is readable. Your failure case is real and worse than it looks: the meter still contains the silent battery's contribution, but the sum that is supposed to remove it no longer has the term. The demand comes out low by exactly that battery's output, and the ceiling trims the healthy one to it.

`None` belongs there as soon as *any* configured battery is unreadable — the quantity is a whole-fleet reconstruction and a partial one isn't a weaker version of it, it's a wrong one.

### 3. A third one, mine, that you haven't hit yet

The deadband shortcut at `__init__.py:8867` needs a pending-check per guard, or the guard never runs. I wrote `_primary_feedforward_pending` and `_surplus_guard_pending` for exactly that reason — the grid reads on-target *because* the other regulator is carrying the load, which is the situation these features exist to change.

There is no `_discharge_ceiling_pending`. So a standing over-ceiling discharge with a quiet meter is skipped — which is precisely the 21:39 case the ceiling was built for, where the EMMA absorbed the excess and the meter sat near zero. The feature is defeated in its own motivating scenario. Worth knowing before you port the logic, because the shortcut will silently swallow anything you add there.

### Two more traps that cost me real time

**One clock, not two.** `read_remaining_solar_kwh` keeps its own `now`. Left alone it converts the solar side against a different minute than the consumption side was trimmed to, and any test that pins the clock only moves half the calculation. Pass `now` in explicitly.

**Scarcity is measured against the DC battery's room, not the fleet's.** I had it the other way first. Asking "is there enough sun to fill everything" is almost never true on a mixed installation, so every day came out scarce and the AC battery was passed over on all of them — while its own empty capacity was most of what made the day look scarce. It sat at its 12 % floor taking **0.02 kWh in a day**. Measured against the DC battery's room instead, the next day it took **13.99 kWh**. The threshold is the whole feature; the ordering around it is comparatively easy.

### The measurements need a home

`site-docs/reference/primary-battery-field-evidence.md` is 292 lines and only exists on this branch. If you close this PR it goes with it, and it's the part you said you can't reproduce — the charge order across two days, the AC/DC conversion table (6.5 % at full load, 27 % at absorption power), telling shade from control via a lux sensor, the absorption tail, and the 21:39 reading you quoted.

Say the word and I'll put it up as a docs-only PR against `release/v1.4.0`, with the implementation claims rewritten as observations so it doesn't describe code that isn't there yet. It's more useful to you as a reference in the tree than as a closed branch.

### The installation stays available

That part is easy to agree to. Marstek Venus D and a SUN2000/LUNA2000 behind an EMMA on one meter, and I'm running your builds on it anyway. Tell me what you want measured and I'll measure it — including against your implementation once it lands, which is the test that actually matters here.

ffunes:
Retargeted to `release/v1.5.0` so it can't land in 1.4.0 by accident. The plan is unchanged: this closes once the split is complete, pointing at #380 and #381.

Yes to the docs-only PR. Put `primary-battery-field-evidence.md` up against `release/v1.5.0` with the implementation claims rewritten as observations. It is the part I can't reproduce and I don't want it going with the branch.

I checked the split rather than taking it on trust: #380's coordinator patch is identical to this branch's, offsets aside, and #381 drops `dc_coupled` and rewords §13.9–13.10 so they describe measurements instead of code that doesn't exist yet. Nothing smuggled either way.

Thanks for checking my two objections against the code, and for the third. All three confirmed on this branch:

- `_uncovered_load_w` returns a number as soon as one coordinator answers, so a silent battery that is still delivering understates the demand by exactly its own output;
- `_apply_surplus_guard` and `_apply_discharge_ceiling` compare against 0, with no reference to `compute_active_target()`;
- there is no `_discharge_ceiling_pending` in the deadband shortcut. That is the one I would have walked into: the ceiling is defeated in the 21:39 case it was built for.

The two traps are noted as well — one clock passed in explicitly rather than two, and scarcity measured against the DC battery's room rather than the fleet's. 0.02 kWh against 13.99 is not a detail, and it is the kind of thing only a real installation says out loud.

I'll build it in `control/` for 1.5.0 and come back with something for you to measure against.

ffunes:
Thank you for this — and for the measurements especially. Everything here is now on `release/v1.5.0`, so I am closing the PR rather than merging it, and I want to be explicit about why.

The Huawei driver, the `drivers/base.py` seam and the two coordinator defects (the zero power ceiling, and a native daily-energy counter zeroing the system totals) went in earlier, in 7914aa6.

The control layer I reimplemented in 216bbed, taking this PR's description as the specification, split into `control/residual_load.py` (the meter side: one `grid + sum(ac_power)` reconstruction feeding the feedforward floor, the surplus guard and the discharge ceiling) and `control/charge_order.py` (share by remaining room, order by the day). Three things came out different in the process, and they are worth naming since they were defects on this branch:

- the guards took `0` as their floor, which vetoed `curtailment_predischarge`'s deliberate negative override under sun and trimmed it to house load at night; they now take `compute_active_target()`;
- `uncovered_load_w` returned `None` as soon as *any* configured battery was unreadable — but a silent battery is still delivering, so its output stays in the meter reading while its term leaves the sum;
- there is one `guards_pending()` derived from the pipeline rather than a flag per guard, wired into both the deadband and the stale-sample early returns, so a guard added later cannot be silently swallowed.

Your tests were replaced by `tests/test_residual_load.py` and `tests/test_charge_order.py`, which carry the same measured numbers against the new structure.

The two documents are the part nothing else replaces, and they are in as ac82c3b under your authorship: the EMMA register map, and the 26–27 August field evidence — both now in the docs nav (the field-evidence page was unlinked here). Yours is still the only installation with that pairing, and the charge order rests on those numbers.

Credited in the changelog for v1.5.0. Your reservation about the forecast-driven half having had the least exposure is noted and carried over.

PR #352, opened 2026-08-28
Let an excluded device claim part of the solar forecast

syphernl:
Implements #341.

Predictive charging counts the whole remaining solar forecast as available to the battery, also when an excluded device the home sensor already sees is going to eat it. The balance then reports "sufficient energy" and skips the cheap grid slots.

A new optional per-device field points at a sensor reporting the energy the device still expects to consume today (evcc publishes one per loadpoint):

```
claim = min(sum(device_remaining_kwh), solar_remaining_effective_kwh)
solar_available_to_battery = solar_remaining_effective_kwh - claim
```

Off unless configured; without it every number is unchanged.

Notes for review:

- Only `included_in_consumption` devices may claim, and the claim is capped at the available solar. Both avoid counting the same energy twice, since anything else is already inside the consumption forecast.
- `exclusion_pct` scales the claim exactly as it scales the consumption correction: at 50 % the consumption forecast still carries half the device's demand, so only the other half is reserved. (An earlier revision of this description said the opposite; `8e2ac38` made the two agree.)
- The chronological planner and the evening/SOC-drop pass deduct the same claim, otherwise they would plan against solar the device is going to take.
- Re-evaluation is a predicate polled from `handle_dynamic_pricing_predictive_charging`, in the shape of the existing evening and SOC-drop triggers, rather than a new state listener. Bidirectional, 2 kWh threshold, 15 min cooldown, 4/day.
- Existing diagnostic keys keep their meaning; the claim is published separately as `excluded_demand_claim_kwh`.
- The learned-profile fingerprint is deliberately untouched, so upgrading does not invalidate anyone's consumption profile.

`pytest`: 2116 passed, 10 skipped. New unit tests cover the deficit chain (including a baseline proving an unconfigured install is unchanged), the sensor reader, the re-evaluation predicate, both config flows, the attribute and the diagnostics payload. `_evaluate_evening_recharge` has no test coverage in the repo today, so the deduction there is untested.

Docs and all six translations updated.


ffunes:
Thanks for the thorough implementation and test coverage. The main scenario works, but I found a few accounting inconsistencies that should be addressed before merging:

1. `claimable_solar_demand_kwh()` does not follow the same accounting rules as `consumption_delta_kw()`. The latter applies `exclusion_pct` and skips `ev_charger_no_telemetry` devices, while the new claim includes their full remaining demand. This double-counts demand in both cases and can schedule unnecessary grid charging. For example, with 6.5 kWh remaining and 50% exclusion, the consumption forecast retains 3.25 kWh while the full 6.5 kWh is also deducted from solar.

2. `_read_sensor_kwh_opt()` treats every unit other than Wh/MWh as kWh. Since the selector accepts any sensor, values in W, J, GJ, or another unsupported unit can reserve the entire solar forecast. Please restrict the selector to energy sensors and safely convert or reject unsupported units.

3. `_evaluate_evening_recharge()` updates `excluded_demand_claim_kwh` but leaves `solar_available_to_battery_kwh` and `solar_remaining_effective_kwh` unchanged. The calculation uses the new claim, while the published diagnostics can retain the morning values.

4. In an extended cross-midnight projection, tomorrow’s forecast is added before the claim is applied as a margin over the combined timeline. A sensor representing demand remaining today can therefore reduce tomorrow’s solar projection.

The full suite passes locally (`1864 passed, 10 skipped`), so these cases need additional regression tests. I think the claim eligibility and weighting should share the same accounting policy as the consumption forecast.

syphernl:
Thanks — all four are addressed, and the branch is rebased onto `release/v1.4.0`.

1. `claimable_solar_demand_kwh()` now mirrors `consumption_delta_kw()` exactly: `exclusion_pct` scales the claim, and `ev_charger_no_telemetry` devices are skipped. At 50 % exclusion the forecast keeps half the demand and only the other half is reserved.
2. The reader converts through Home Assistant's `EnergyConverter` and claims nothing when the unit is not an energy unit, so a sensor reporting W or J can no longer reserve the forecast. The match is exact on purpose: `mWh` and `MWh` differ only in case. The config-flow selector is restricted to energy sensors as well.
3. The evening re-evaluation now publishes the figures that horizon actually used, including the zero safety margin it applies, instead of leaving the morning values in place.
4. The claim is no longer folded into the timeline's safety margin. `_apply_excluded_demand_claim` takes the reservation out of today's intervals only, proportional to their energy, and caps it at what those intervals hold. Tomorrow's forecast in a cross-midnight projection is never touched; for a same-day horizon the result is unchanged.

Regression tests added for each, including the eligibility parity with `consumption_delta_kw`, the unit rejection, and a cross-midnight case.

While testing point 3 I hit the `get_average` NameError in the evening path. You had already fixed it in e8f468f, so the rebase dropped my duplicate.

Two more things, in a separate `test:` commit you can drop if you would rather fix them yourself. Both reproduce on a pristine `release/v1.4.0` checkout and are independent of this branch, but they keep `pytest -o addopts=""` red:

- `31 passed, 1865 errors — RuntimeError: There is no current event loop`. Unit tests that drive coroutines with `asyncio.run` close the loop and clear the thread's current loop, so with the Home Assistant plugin active every following test errors in its fixtures. An autouse fixture now restores a fresh loop after a test that left none.
- `test_balance_needs_charge_unlocks_low_forecast` reads the wall clock and asserts `low_forecast`; after T_end the reason is `past_t_end`, so it fails every evening. It now pins `_decision_now` to 10:00, the idiom the neighbouring tests already use.

With both, `pytest -o addopts=""` reports 1963 passed, and CI on this PR is green.

ffunes (changes requested):
Nice piece of work — tests, docs and all six translations in one go.

I checked the load-bearing premise rather than taking it on trust: `consumption_tracker.py` feeds `consumption_delta_kw()` into the learned profile, and `external_loads.py` subtracts `device_kw * _exclusion_factor(device)`. So an `included_in_consumption` device's demand is already out of `avg_consumption_kwh` at exactly the `exclusion_pct` weight, and reserving the same fraction from solar is not a double count. The eligibility filters in `claimable_solar_demand_kwh` mirror `consumption_delta_kw` one for one. I also confirmed `configuration_fingerprint` uses an explicit field allowlist, so the new key genuinely cannot invalidate anyone's learned profile, and translation parity is complete. Full suite on the branch here: **1953 passed, 10 skipped**.

**The design is approved — nothing below asks you to rework it.** Three things before merge:

**1. The PR description contradicts the code.** It says "`exclusion_pct` is not applied: it governs live power, not a future energy demand", but the code, the docstring, the docs and the tests all apply it. Stale after `8e2ac38`. This is the text that lands on #341, so please update it.

**2. The field is silently ignored for `ev_charger_no_telemetry` devices** — which is exactly the user who has an evcc remaining-energy sensor and *no* wattmeter. The skip itself is correct (their energy stays inside the consumption forecast, so a claim would double count), but the form offers the field and the `data_description` only names the `included_in_consumption` condition. Please add the caveat to both `data_description` entries across the six locales, or reject the combination in the flow. You have the evcc setup, so you're better placed than me to word it.

**3. A transient unavailable resets the trigger reference to zero.** `_refresh_excluded_demand_reference` does `0.0 if current is None`. If the sensor blips during an evening or pre-slot refresh the reference drops to 0, and its return at 20 kWh then reads as a +20 kWh move and fires a spurious re-plan. Keeping the old reference is strictly better, since `_is_excluded_demand_reeval` already refuses to fire while `current is None`, so nothing is missed:

```python
if current is not None:
    self._controller._dp_last_eval_excluded_claim_kwh = current
```

Worth folding into the same push, but not blocking on their own:

**4.** `_refresh_excluded_demand_reference`'s "Called from every path that replaces `_last_decision_data`" is not true — 3 of 12 assignment sites call it. Harmless today, since the cooldown and the daily cap bound the fallout, but it will mislead the next reader.

**5.** "Spread evenly" in the `_build_chronological_plan_for_horizon` comment and in `excluded-devices.md` contradicts `_apply_excluded_demand_claim`, which scales proportionally to interval energy — as its own docstring correctly says. Proportional is the right behaviour; only the two places that say otherwise need fixing.

**6.** Minor: the selector `EntitySelectorConfig(domain="sensor", device_class="energy")` is stricter than `_read_sensor_kwh_opt`, which only needs a convertible unit. A template helper without `device_class: energy` can't be picked even though the runtime would accept it. Your call.

One question, no change required unless you think it needs one:

**7.** Phase 2.7 runs a full `_evaluate_dynamic_pricing(REMAINING)` with no active-slot guard. `_check_dp_pre_slot_reevaluation` deliberately returns early on `_current_price_slot_active`, and the #411 SOC-drop trigger uses the lighter `_evaluate_evening_recharge`. This is the only path that rebuilds the whole schedule on top of a live charge, so an EV session ending mid-slot can shrink the plan and drop the running slot. If that's intentional, a comment saying so is enough.

Finally: the `conftest.py` event-loop fixture and the `test_charge_delay` clock pin are unrelated to this feature, but they fix real flakes and splitting them costs more than it's worth — **keep them**. I'll open a separate issue for the fixture's `policy.get_event_loop()`, which emits a `DeprecationWarning` and is removed in Python 3.14; don't hold this PR for it.

syphernl:
@ffunes Done!

PR #355, opened 2026-08-28
feat(predictive): let PV surplus wait for the cheapest feed-in hours

syphernl:
Implements the behaviour asked for in discussions #237 and #346. Both describe the same thing: on a dynamic contract, storing PV surplus forfeits that slot's feed-in revenue, so the same daily charge is cheaper taken in the lowest-priced hours. #237 additionally asks for a separate export price sensor, which is a config input to the same planner, so this is one opt-in feature rather than two.

**What it does.** A pure planner (`pricing/surplus_absorption.py`, shaped like `pricing/curtailment.py`) derives the remaining target from consumption, stored energy, the solar-forecast safety margin and free space, then picks the cheapest export hours that cover it, accumulating energy rather than hours. A thin runtime (`control/surplus_price_hold.py`) registers a global `surplus_price_hold` charge blocker outside those windows: the charge pool empties, PD clamps to 0 W and the surplus exports. Discharge is untouched.

It releases inside a selected window, past the end of solar production, once the target is met, when the remaining windows can no longer cover it, and when the best remaining slot is not cheaper than a configurable minimum saving. The target is re-derived from live battery state every cycle (snapshots only, no recorder work in the control loop), so an under-delivering cheap window releases the hold rather than ending the day short.

Charge delay (including its per-battery setpoint phase), scheduled cheap grid slots, negative-price charging, smart pre-discharge, weekly full charge, peak shaving, EV pause, manual ownership and the SOC floor all take precedence. Every missing or unusable input fails safe to absorbing.

**Export price sensor.** Optional, with its own provider type, falling back to the import curve when unset. It reuses the import parsers via two extractions (`calculations.parse_prices_for_integration`, `PricingManager._parse_sensor_price_slots`) and never writes `_price_data_status`, so a flaky export sensor cannot raise the import-price repair issue. Tibber is rejected there because its cache belongs to the import curve.

Adds a `surplus_price_hold_status` binary sensor, an Integration Status state, a runtime minimum-saving slider, diagnostics, docs and translations. Disabled by default. Full suite: 1916 passed, 10 skipped.

**Open questions for you**

- The absorption margin reuses `CONF_PREDICTIVE_SAFETY_MARGIN_KWH`. It is documented as a solar-forecast margin, so say the word if you'd rather have a dedicated constant.
- `CONF_NEGATIVE_INJECTION_THRESHOLD` still reads the import curve. Switching smart pre-discharge to the export curve once one is configured would change behaviour for existing users, so I left it alone.
- Deferred from #346: grid-charging in the cheapest remaining hours before the evening peak when surplus fell short. That belongs in `_evaluate_evening_recharge` / the DP deficit rather than a parallel charging path.

**Unrelated, spotted while working here:** `_evaluate_evening_recharge` in `pricing/engine.py` calls `get_average()`, which is not defined in that scope (it is a local of `_evaluate_remaining_grid_charging`). It looks like it would raise `NameError` on every evening re-evaluation since 8f4e542. Not touched in this PR.


ffunes (changes requested):
Thanks Frank — this is a well-shaped PR. The planner/runtime split mirrors `curtailment.py`, the fail-safe-to-absorbing default is the right bias, and the precedence list against charge delay / scheduled slots / weekly full charge is complete as far as I could check. I read every hunk and opened the surrounding code (`_should_activate_grid_charging`, `_curtailment_battery_snapshots`, `distribute_solar_forecast`, the blocker registry, the ordering in `_run_control_cycle`) rather than just the diff.

Three things to fix before merge, one of them blocking.

---

### 1. Blocking — the plan target and the live target measure different energy, so the hold never engages on most configurations

`custom_components/omnibattery/control/surplus_price_hold.py:237` vs `:306`

The two paths feed `calculate_absorption_target_kwh` two different definitions of `usable_energy_kwh`:

- **Plan time** takes `decision["usable_energy_kwh"]` from `_should_activate_grid_charging` (`__init__.py:4482`): *all* `coordinators_with_data`, measured against `coordinator.min_soc` — the hardware cutoff.
- **Every cycle**, `_refresh_live_target` recomputes it via `_usable_energy_kwh` (`:314`): *eligible* batteries only, measured against `snapshot.floor_soc_pct`, which `_curtailment_battery_snapshots` (`pricing/engine.py:1474`) sets to `max(_effective_discharge_min_soc, guaranteed floor, predischarge_reserve_soc)`.

The live figure is the lower of the two whenever the discharge min SOC sits above the hardware cutoff, a pre-discharge reserve or guaranteed floor is set, or any battery is ineligible — i.e. most real setups. Since `needed = consumption + margin - usable`, a lower usable means a *higher* live target, while `select_absorption_slots` only picked enough slots to cover the *lower* plan target. So on the first cycle after every rebuild:

```
plan.remaining_absorbable_kwh(now)  ≈ plan.target_kwh  <  live_target_kwh
→ hold_decision() hits `remaining + EPSILON < target`
→ returns (False, REASON_SHORTFALL_RISK)
```

The feature reports `shortfall_risk` and silently never holds. `calculate_free_space_kwh` runs off the same snapshots in both paths, so this one term is the only thing that flips.

Which definition you want is your call, and it's a real design choice rather than a typo:

- Feeding the snapshot-based figure into the plan too keeps one definition and makes the reserve genuinely off-limits to the absorption plan — but it sizes the target larger, so the hold reserves more of the day's cheap windows.
- Keeping the `_should_activate_grid_charging` figure on both sides matches what the rest of predictive charging already means by "usable", at the cost of the plan treating reserved energy as available.

Either is defensible; what can't stand is one on each side. Whichever you pick, a test that sets `predischarge_reserve_soc` (or marks one battery ineligible) and asserts the hold still engages on the cycle after a rebuild would have caught this, and none of the current cases exercise it — every test I found leaves the floor equal to the hardware min.

### 2. The hold stays latched for the whole of a manual session

`custom_components/omnibattery/__init__.py:7915`

The global manual-mode early return fires before `_refresh_operation_blockers()` (line 7945), so an active `surplus_price_hold` charge blocker and its `holding` status survive untouched until manual mode ends — `_release_guard`'s `GUARD_MANUAL` branch is never reached, because nothing calls into the manager at all. Line 7916 right above already does exactly this for the curtailment runtime:

```python
if self.manual_mode_enabled:
    self._pricing_mgr.clear_curtailment_runtime("manual_mode")
+   self._surplus_hold_mgr.clear("manual_mode")
```

### 3. `SurplusPriceHoldSensor` registers without a device

`custom_components/omnibattery/binary_sensor.py:238`

It's the only system-level binary sensor in the file with no `device_info` property (cf. lines 91, 175, 230, 422, 809), so it lands outside the "Omnibattery System" device and is missing from the device page and any device-scoped grouping. Same four-line block as its neighbours.

---

### Minor, take or leave

- `_get_current_export_price` and the `build_absorption_plan` alias have no production callers — only tests. The alias mirrors `build_curtailment_plan`, so I don't mind it, but worth knowing they're both dead outside the suite.
- `_release_guard`: `BLOCKER_SOURCE != "ev_pause"` is constant-true.

### Your open questions

- **Reusing `CONF_PREDICTIVE_SAFETY_MARGIN_KWH`:** keep it. A dedicated constant is another option to document, translate and explain; add one when someone actually wants the two margins to differ.
- **`CONF_NEGATIVE_INJECTION_THRESHOLD` on the import curve:** agreed, leave it. Switching it silently would change behaviour for every existing user with a configured export sensor.
- **The #346 grid-charging deferral:** agreed, that belongs in `_evaluate_evening_recharge` / the DP deficit, not a parallel charging path. Out of scope here.

### The `get_average()` NameError

Good catch, and it's real — confirmed it's a local of `_evaluate_remaining_grid_charging`. It predates this branch, so please leave it out of this PR; I'll open a separate issue.

ffunes:
Correction on my last point: no separate issue needed after all.

The `get_average()` `NameError` was real at your merge base (`c9eee90`), so your reading of it was right — but it was already fixed on `release/v1.5.0` by e8f468f ("fix: resolve evening pricing consumption average", 28 Aug), which swaps the stray local for `await tracker.get_dynamic_base_consumption()` and adds a regression test that executes the evening path. Your branch is just behind the base.

So nothing to do here beyond a rebase, which you'll want anyway before the three fixes above. Sorry for the misdirection.

syphernl:
@ffunes PR has been rebased.

ffunes:
Re-reviewed at `88fed54`. Thanks for the rebase — it does pick up `e8f468f`, so the `get_average()` question is closed. Everything else from my earlier review is still in the tree; I checked each one against the current head rather than trusting my old line numbers. Two of the three moved, none changed.

### Still open

**1. Blocking — the plan target and the live target measure different energy.** `control/surplus_price_hold.py:237` still passes `decision.get("usable_energy_kwh")` into the planner, while `:306` recomputes it via `_usable_energy_kwh(snapshots)` at `:314`. Plan side: all `coordinators_with_data` against `coordinator.min_soc`. Live side: eligible batteries only, against `snapshot.floor_soc_pct`. Any install with a pre-discharge reserve, a guaranteed floor above the hardware cutoff, or one ineligible battery gets a higher live target than the plan bought slots for, so `hold_decision()` returns `shortfall_risk` on the first cycle after every rebuild and the hold never engages. Details and the two defensible ways to resolve it are in my earlier review — either definition works, one on each side doesn't.

**2. The hold stays latched through a manual session.** Now at `__init__.py:8159` (was 7915). The global manual-mode early return still clears the curtailment runtime and returns before `_refresh_operation_blockers()`, so the `surplus_price_hold` charge blocker and its `holding` status survive untouched. Still the same one-line fix next to its neighbour:

```python
if self.manual_mode_enabled:
    self._pricing_mgr.clear_curtailment_runtime("manual_mode")
+   self._surplus_hold_mgr.clear("manual_mode")
```

**3. `SurplusPriceHoldSensor` has no `device_info`.** `binary_sensor.py:240`. Still the only system-level binary sensor in the file without it.

Minors unchanged: `_get_current_export_price` has no production caller (only `tests/test_export_price_curve.py:181`), and `BLOCKER_SOURCE != "ev_pause"` at `surplus_price_hold.py:486` is constant-true.

### Two more, spotted this pass

**4. The five non-English translations carry English text.** `ca/de/es/fr/nl` all have the new keys present — `surplus_price_hold_enabled`, `surplus_hold_min_saving` and their `data_description`s — but the values are the English strings verbatim. Every neighbouring key is properly localized in all five (compare `negative_price_charging_enabled`), so this reads as an oversight rather than the repo's convention. Nothing breaks; a German user just gets half a config step in English. Happy to write these five myself in a follow-up commit if you'd rather not — say the word.

**5. The dashboard panel never renders the new sensor.** `frontend/marstek-panel.js` isn't touched by this PR. Its sibling `curtailment_status` needs five places in that file to appear: a key in `K` (line ~887), a row in `DIAG_ROWS` (~907), an entry in the dynamic-pricing card's `items:` allowlist (~1128), a tone `case` (~4658), and a label in each of the six inlined language maps. `surplus_price_hold_status` is in none of them, so the entity exists but is invisible on the dashboard — which is where a user would go to see whether the hold is on.

The branch is two commits behind `release/v1.5.0` (`1cc7c74` vs `a50c844`); no conflict, just noting it.

ffunes:
Merged. Re-checked every point against `2911e03` and all five are properly closed — the target alignment via a single `calculate_usable_energy_kwh()` on both sides is the right call, and the three bugs you found along the way (the `_daylight_share` midnight deadline, the `plan.slots` desync, the health-status restore masking real sensor faults) were worth the round trip on their own.

Two things I fixed on the way in rather than sending back, both in `4f579cf`:

- **The export price fields were still English in ca/de/es/fr/nl.** Same class as the two hold keys you did localize — `export_price_sensor` and `export_price_integration_type`, label and `data_description`, both in the config and options steps.
- **Trimmed the changelog entry.** House style under `[Unreleased]` is a one-liner; yours ran about ten times its neighbours. The full description belongs in the PR, which is where it now lives.

The `CHANGELOG.md` conflict was just your branch predating `94554e7`; entry moved to `[Unreleased] / Added` untouched. Thanks Frank.

PR #362, opened 2026-08-29
Feature/fronius gen24 byd

MisterSpliss:
## Summary

This PR adds Fronius GEN24 with BYD battery support and automatically detects both supported SunSpec register layouts:

- `float`
- `int+SF`

The detected SunSpec model type is shown in the battery driver information panel.

## Changes

- Added the Fronius GEN24/BYD driver.
- Added automatic SunSpec Model 124 layout detection.
- Uses the correct register addresses for the detected layout.
- Added atomic FC16 writes for the adjacent `OutWRte` and `InWRte` registers.
- Added a safe `0/0` idle power window.
- Added a short idle period when changing between charging and discharging.
- Added delivered-power confirmation after setpoint changes.
- Added a persistent switch controlling Fronius ownership:
  - Enabled: Omnibattery retains external control.
  - Disabled: control is explicitly returned to the Fronius inverter.
- Returning control to Fronius only resets `StorCtl_Mod`; it does not modify `MinRsvPct` or the configured power windows.
- Batteries released to Fronius are excluded from all Omnibattery automatic control, pricing, charge-delay, weekly-charge, balancing, and power-distribution paths.
- Restored Config Entry version 12 compatibility and added migration from version 11.
- Added diagnostics for the detected layout and current ownership state.
- Updated the dashboard, documentation, changelog, and all included translations.

## Safety behavior

The default behavior retains external control and places the battery in a closed `0/0` idle window during setup, reload, and shutdown.

Returning the battery to Fronius automatic control requires an explicit action using the new device switch. This choice is persisted across Home Assistant restarts.

## Testing

- Successfully tested on a real Fronius GEN24 system with a BYD battery.
- Automatic SunSpec model detection worked correctly.
- Home Assistant loaded the existing version 12 Config Entry successfully.
- Battery telemetry and control were verified on the hardware.
- The detected SunSpec model type was displayed correctly in the driver information panel.
- All 25 Fronius driver tests passed.
- Python and JavaScript syntax checks passed.
- Translation JSON validation passed.
- `git diff --check` passed.


ffunes:
Thank you for the substantial work on adding Fronius GEN24 / BYD support. The driver structure, telemetry coverage, ownership switch, translations, documentation, and dedicated unit tests provide a strong foundation.

We reviewed PR #362 against its intended target, `release/v1.4.0`. We cannot merge it yet because:

- The branch needs to be rebased or updated against `release/v1.4.0`, preserving the fixes already present there. It currently conflicts in several central files, including integration setup, config flow, coordinator, ownership handling, switches, and control logic.
- Releasing ownership back to Fronius may still allow a persisted software-manual Charge/Discharge setpoint to be reapplied. Releasing the battery should clear or suspend those commands and guarantee that Omnibattery sends no further setpoints.
- The power-window writes always encode percentages using a denominator of 10000, although the device-provided `InOutWRte_SF` is read. The actual scale factor should be used when encoding commands, or unsupported values should be rejected safely.
- SunSpec Models 160 and 124 are detected only at two fixed address pairs. Fronius devices can have different model positions depending on their active model list and MPPT blocks. We need proper model discovery or a clearly documented and tested compatibility restriction.
- Capacity reported by the Fronius API is currently overwritten by the configured default because the driver declares that it has no nominal-capacity telemetry. We need one clear source of truth and a coordinator-level regression test.
- The normal test configuration disables the Home Assistant pytest plugin and skips tests requiring the `hass` fixture. Please add a Linux CI run with the plugin enabled, covering setup, config flows, migration, ownership transitions, and unload/reload.
- `ac_power` is currently derived from DC battery power; its semantics should be validated with end-to-end power-balance tests.
- Please remove the accidental `frontend/._marstek-panel.js` AppleDouble file and complete the general driver/support documentation.

The 25 dedicated Fronius unit tests pass, which is a good starting point. Once these issues are addressed and CI is green, we will be happy to review the updated version. Thanks again for contributing Fronius GEN24 / BYD support to Omnibattery.

PR #380, opened 2026-09-01
Two coordinator defects in device-reported values

sphings79:
Split out of #335 as you asked. Two coordinator fixes, one file plus its tests, no control-layer code and no dependency on the rest of that PR.

### A reported ceiling of zero is a missing answer, not a limit

A Marstek came back from a restart with both power registers reading 0. Adopting that shut the battery out of every allocation — it could neither charge nor discharge — and nothing would ever write those registers again, because the controller had stopped addressing a battery it believed could do nothing. Only a manual write to the number entity brought it back.

The last known ceiling now stands when the device reports 0, and the mismatch gets a warning that says which entity restores it.

### The daily reset stamp

The system totals refuse to add up unless every battery's daily figure is marked as belonging to today — a guard against summing one battery's fresh value with another's from before midnight. Only the derived counter stamped that mark, because until now every driver needed one. A driver whose device keeps its own daily counters never gets that sensor, so it never carried the mark, and one such battery in a fleet zeroed the system totals outright: **10.3 kWh charged on one battery and 3.97 on the other, with the overview reading 0.00.**

Stamping every poll as today was the obvious fix, and I shipped it first. It was wrong. A device keeps its own midnight, and for a few minutes either side of ours its counter still holds yesterday's accumulation. Marking that as today let the aggregate add a battery's fresh 0.00 to another's stale 14.13 and latch the sum — the aggregate refuses same-day decreases, so the wrong figure stood until the real total grew past it, which on the observed day was never.

The mark now waits for evidence that the device's own counter has turned: the value dropping below what was last seen. Until then the previous day's mark stays and the aggregate treats the sum as incomplete. Better no total than a wrong one that sticks.

A counter last seen at zero carries nothing stale, so it needs no such evidence. The function is gated on `has_daily_energy_counters`, which is itself the claim that the device resets daily.

### Tests

12 new tests in `tests/test_coordinator_device_reports.py` — the zero-ceiling cases for soft-max and writable-register drivers, and four midnight-transition cases that keep state across polls: our midnight before the device's, the device's counter turning, a counter last seen at zero, and a restart with no history.

### One thing you should know about the base

`tests/test_charge_delay.py::test_balance_needs_charge_unlocks_low_forecast` fails on current `release/v1.4.0` before this branch touches anything:

```
assert ctrl._charge_delay_status["unlock_reason"] == "low_forecast"
AssertionError: assert 'energy_balance' == 'low_forecast'
```

I verified it by stashing this change and running the test on the clean base. Not mine, and not fixed here — flagging it rather than folding an unrelated fix into a PR you want small.

Full suite on this branch: 1930 passed, 10 skipped, that one pre-existing failure.

### DIFF
--- a/custom_components/omnibattery/infra/coordinator.py
+++ b/custom_components/omnibattery/infra/coordinator.py
@@ -64,6 +64,121 @@ def _normalise_power_limit(value) -> int:
         return 0
 
 
+def _stamp_native_daily_reset_dates(coordinator) -> None:
+    """Date-stamp daily energy values a device counts for itself.
+
+    The system totals refuse to add up unless every battery's daily figure is
+    marked as belonging to today — a guard against summing one battery's fresh
+    value with another's from before midnight. Only the derived counter stamps
+    that mark, because until now every driver needed one.
+
+    A driver whose device keeps its own daily counters never gets that sensor,
+    so it never carried the mark, and one such battery in a fleet zeroed the
+    system totals outright: 10.3 kWh charged on one battery and 3.97 on the
+    other, with the overview reading 0.00.
+
+    The poll happening today does not make the *value* today's. A device keeps
+    its own midnight, and for a few minutes either side of ours its counter
+    still holds yesterday's accumulation. Marking that as today let the system
+    aggregate add a battery's fresh 0.00 to another's stale 14.13 and latch the
+    sum: the aggregate refuses same-day decreases, so the wrong figure stood
+    until the real total grew past it — all day, in the observed case.
+
+    So the mark waits for evidence that the device's own counter has turned,
+    which is the value dropping below what was last seen. Until then the
+    previous day's mark stays and the aggregate treats the sum as incomplete,
+    which is the honest answer: better no total than a wrong one that sticks.
+
+    A counter last seen at zero carries nothing stale, so it needs no such
+    evidence. This assumes the device does reset daily — the capability that
+    gates this function is the claim that it does.
+    """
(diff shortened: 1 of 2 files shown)


ffunes:
Reopening to pick up the fixed base — see the comment that follows.

ffunes:
Both base problems you flagged are fixed on `release/v1.4.0` and merged forward, so this now sits on a green base. Retargeted to `release/v1.5.0`, and closed/reopened only to make the checks re-run against it.

- The CI break was mine (ad0b54e). `pytest -o addopts=""` drops the `-p no:homeassistant` that `pytest.ini` sets, which loads the HA plugin for all ~1900 tests; its autouse `enable_event_loop_debug` calls `asyncio.get_event_loop()`, and a sync test has no loop. Your diagnosis was exact. Fixed by splitting the run rather than pinning versions — pinning would only have delayed the same collision, since the plugin assumes an all-async suite and this one isn't. Unit suite with the plugin off; the eleven `hass`-based tests (domain migration, config backup, system-uid heal) as their own step with it on, so ad0b54e's intent survives. The workflow now also runs on `release/**`, which is why nobody saw the base go red until every PR did.
- `test_balance_needs_charge_unlocks_low_forecast` was clock-dependent, as you suspected. Consumption is trimmed to the hours left in the day, so the deficit its own comment documents only holds before midday and it failed on any afternoon run. Clock pinned at the full-day horizon.

1.4.0 is finishing its beta cycle, so nothing from the split lands there — 1.5.0 is where this goes. Merging as soon as the checks come back.

ffunes:
Thanks a lot for your contribution and work!

PR #381, opened 2026-09-01
Huawei: release instead of commanding while the strings are lit

sphings79:
Second of the split from #335. Huawei only — no control-layer code, and `dc_coupled` stays behind with the rest of that work.

You asked for two things here. The first I agree with and have done: the behaviour change is now the point of the PR rather than an undescribed side effect. The second I want to argue with before implementing, because the measurement that produced this gate is already in the tree and it says the opposite.

### What the change is

`apply_setpoint` drops to 0 whenever the strings carry voltage. That is a driver-wide behaviour change and it deserves the headline, so here it is in full.

**A forcible command on this hybrid is a ceiling, not a request.** The inverter produces exactly what it was told and curtails the rest of the array. Caught with a 315 W charge standing: **288 W harvested from strings that made 5054 W six seconds after the command ended**, while a separate balcony array on the same roof held steady throughout — so not weather.

A forcible discharge is worse, and this is the part that matters for your objection. It serves the house from the battery and leaves the MPPT tracker down entirely: **374 V at 0.00 A**. With the roof producing nothing, the meter shows a real deficit — which is what asks for discharge. The command goes on justifying itself. **Commanded at 04:32 against a genuinely dark sky, still standing at 09:22.**

So while there is light on the strings this driver commands nothing and hands back to the inverter's own regulation. On a DC-coupled hybrid the inverter is the better controller in daylight: it is the only party that knows what the array could be making. The cost is real and stated in §13.8 — a Huawei battery is under Omnibattery's control after dark only.

### Why I have not gated it on harvested power

Your dawn case is real. At 300 V and 50 W the gate opens and a discharge the house needs is refused. I want that fixed.

But gating on harvested power reintroduces the 04:32→09:22 bug, because our own command destroys the signal:

| State | String voltage | Harvested |
|---|---|---|
| Night | 0 V | 0 W |
| Overcast dawn | ~300 V | ~50 W |
| **Sun up, curtailed by our own discharge** | **374 V** | **~0 W** |

The last two rows are indistinguishable by harvested power. A discharge standing through sunrise holds the tracker down, `solar_power` stays near zero, the gate never closes, and the command survives the whole day — which is exactly what was observed before the voltage test went in. Voltage is the only signal that survives our own curtailment.

### What I think the right gate is

The driver knows whether a command is standing (`_last_written_w`), and that is what makes the harvested reading trustworthy or not:

- **While released** — which is the steady state under sun, since the gate releases — `solar_power` (register 32064) is the true harvest, and a threshold on it closes your dawn hole cleanly.
- **While a command stands**, harvested power is our own artefact and must not be read. The verdict has to latch, with a short periodic release to re-sample.

That is a small state machine rather than a one-line change, and it is your driver's direction, so I would rather have your call than guess it. I have the hardware to measure whatever threshold you want, across as many dawns as it takes.

Say which way and I will push it onto this branch: the probe design above, a plain harvested-power gate if you have a reason to think the curtailment case is narrower than I measured, or something else you prefer.

### Also in here

- The EMMA register map (`site-docs/reference/huawei-emma-registers.md`), read off the reference installation.
- §13.8–13.10 of the driver assessment: the curtailment finding, what a second battery looks like to the hybrid's energy manager (1391 W of PV over a 529 W house, one battery taking in 1110 W while the other gave up 205 W, meter at 3 W), and the form-schema serialisation failure that made the config flow unreachable for this brand.

### Tests and base

190 pass in `tests/test_huawei_driver.py`, including lit and dark string fixtures. Full suite 1925 passed, 10 skipped.

The one failure is `tests/test_charge_delay.py::test_balance_needs_charge_unlocks_low_forecast`, which fails on clean `release/v1.4.0` before this branch touches anything — `assert 'energy_balance' == 'low_forecast'`. Same note as on #380; not mine, not fixed here.

### DIFF
--- a/custom_components/omnibattery/drivers/huawei.py
+++ b/custom_components/omnibattery/drivers/huawei.py
@@ -112,6 +112,13 @@ _PV_LIT_VOLTAGE_V = 100.0
 # below 100 V, so if a second installation disagrees, these numbers are where
 # to look first rather than the logic around them.
 _PV_HARVEST_THRESHOLD_W = 150.0
+# A band either side of it, because the dawn ramp crawls across the threshold
+# rather than stepping over it. Measured 2 September on a cloudless morning:
+# twenty minutes wobbling between 123 W and 214 W around the 150 W mark, with
+# irradiance rising smoothly throughout — so the wobble is the array, not the
+# weather. A bare comparison would flip the gate on every one of those, and
+# each flip costs a write the inverter answers by derating.
+_PV_HARVEST_HYSTERESIS_W = 50.0
 # How often a held verdict is re-checked by letting go for a moment. The cost
 # is a slice of command time each interval; the alternative is missing sunrise.
 _PV_PROBE_INTERVAL_S = 120.0
@@ -1015,7 +1022,11 @@ class HuaweiSolarDriver(BatteryDriver):
             # No reading is not evidence either way; the last verdict stands.
             return
 
-        if float(harvest) > _PV_HARVEST_THRESHOLD_W:
+        # Latching, with the band above on the way in and below on the way out.
+        threshold = _PV_HARVEST_THRESHOLD_W + (
+            -_PV_HARVEST_HYSTERESIS_W if self._pv_lit else _PV_HARVEST_HYSTERESIS_W
+        )
+        if float(harvest) > threshold:
             self._pv_lit = True
             self._pv_verdict_monotonic = now
             self._pv_probe_until = 0.0
--- a/custom_components/omnibattery/drivers/huawei.py
+++ b/custom_components/omnibattery/drivers/huawei.py
@@ -106,6 +106,11 @@ _PV_LIT_VOLTAGE_V = 100.0
 # while the strings already stand near 300 V — time the battery is needlessly
 # refused. Set low deliberately: too high curtails a roof that is genuinely
 # producing, too low only costs availability.
+#
+# Both this and the voltage threshold above come from a single installation —
+# one ~374 V array on a SUN2000-8K. A smaller string can genuinely produce
+# below 100 V, so if a second installation disagrees, these numbers are where
+# to look first rather than the logic around them.
 _PV_HARVEST_THRESHOLD_W = 150.0
 # How often a held verdict is re-checked by letting go for a moment. The cost
 # is a slice of command time each interval; the alternative is missing sunrise.
@@ -1158,6 +1163,16 @@ class HuaweiSolarDriver(BatteryDriver):
             # cycle, so they must keep being polled even with their entities off.
             "inverter_max_power", "inverter_ac_power", "battery_power",
             "charging_cutoff_capacity", "discharging_cutoff_capacity",
+            # Inputs of the PV gate (see _update_pv_gate). Without these the
+            # gate cannot form a verdict at all: the harvest reads as absent, so
+            # the verdict stays at its initial "not lit" and the driver commands
+            # through full sun — the failure the gate exists to prevent, put
+            # back by an entity toggle. Solar Power is user-facing, and the
+            # string voltages are enabled_by_default False and only arrive
+            # dragged along by mppt{n}_power, so there are two switches between
+            # this gate and its evidence.
+            "solar_power",
+            *(f"pv{index}_voltage" for index in range(1, _MAX_PV_STRINGS + 1)),
         })
 
     # --- concrete methods the coordinator calls without isinstance guards ----
--- a/custom_components/omnibattery/drivers/huawei.py
+++ b/custom_components/omnibattery/drivers/huawei.py
@@ -100,6 +100,23 @@ _TARGET_MODE_TIME = 0
 # clock. Measured on the reference installation: 374 V under sun, 0 V at night.
 _PV_LIT_VOLTAGE_V = 100.0
 
+# Voltage alone cannot tell a bright morning from a dim one, so it is paired
+# with harvested power (32064). Measured across three dawns on the reference
+# installation, the array sits below this for 45-50 minutes after first light
+# while the strings already stand near 300 V — time the battery is needlessly
+# refused. Set low deliberately: too high curtails a roof that is genuinely
+# producing, too low only costs availability.
+_PV_HARVEST_THRESHOLD_W = 150.0
+# How often a held verdict is re-checked by letting go for a moment. The cost
+# is a slice of command time each interval; the alternative is missing sunrise.
+_PV_PROBE_INTERVAL_S = 120.0
+# How long a release must stand before the reading means anything. The tracker
+# recovered in 6 s when measured (288 W to 5054 W); this leaves margin.
+_PV_PROBE_SETTLE_S = 15.0
+# When to give up on a probe whose release never reached the hardware. Longer
+# than the settle plus _MIN_WRITE_INTERVAL_S, which throttles the release write.
+_PV_PROBE_WINDOW_S = 45.0
+
 # Slave ids worth trying when scanning. 1 is the factory default for a direct
 # connection; the rest are what dongles and energy managers hand out. Kept short
 # because each one costs a connection, and the inverter needs 1.5 s of silence
@@ -422,8 +439,14 @@ class HuaweiSolarDriver(BatteryDriver):
         # Which pack slots are populated, learned from the packs that answer.
         # Empty until one has, so a failed read never hides a pack that exists.
         self._packs: set[int] = set()
-        # Strings lit but not harvested — see read_telemetry.
+        # Whether the roof is producing enough to leave the inverter alone.
+        # See _update_pv_gate for why that takes two signals rather than one.
         self._pv_lit = False
+        self._strings_lit = False
+        # Monotonic deadline of a probe in progress, 0.0 when none is.
+        self._pv_probe_until = 0.0
+        # When the verdict last came from a reading this driver could trust.
+        self._pv_verdict_monotonic = 0.0
         self._serial: Optional[str] = None
         # Last command actually written, so the deadband can compare against what
         # the hardware was told rather than against what it currently delivers.
@@ -664,16 +687,8 @@ class HuaweiSolarDriver(BatteryDriver):
         if storage is not None:
             self._storage_model = _STORAGE_MODELS.get(int(storage)) or self._storage_model
 
-        # Is there light on the panels? See _pv_lit for why that decides whether
-        # this driver may command anything at all.
-        voltages = [
-            snapshot.get(f"pv{index}_voltage")
-            for index in range(1, self._pv_strings + 1)
-        ]
-        if any(volts is not None for volts in voltages):
-            self._pv_lit = any(
-                volts is not None and volts > _PV_LIT_VOLTAGE_V for volts in voltages
-            )
+        # Whether this driver may command at all. See _update_pv_gate.
+        self._update_pv_gate(snapshot)
 
         for index in (1, 2, 3):
             if snapshot.get(f"pack{index}_serial_number") or snapshot.get(
@@ -768,7 +783,7 @@ class HuaweiSolarDriver(BatteryDriver):
             min(self._ceiling("charge"), int(net_power_w)),
         )
 
-        if applied != 0 and self._pv_lit:
+        if applied != 0 and self._hands_off():
             # A forcible command on this hybrid is not a request but a ceiling:
             # the inverter produces exactly what the command asks for and
             # curtails the rest of the roof. Measured with a 315 W charge
@@ -935,6 +950,103 @@ class HuaweiSolarDriver(BatteryDriver):
             battery_power_w=int(battery_power) if battery_power is not None else None,
             applied=echo,
         )
+
+    def _hands_off(self) -> bool:
+        """Whether the inverter is to be left to its own regulation right now."""
+        return self._pv_lit or bool(self._pv_probe_until)
+
+    def _update_pv_gate(self, snapshot: dict) -> None:
+        """Decide whether the roof is producing enough to leave alone.
+
+        Two signals, and neither is sufficient by itself.
+
+        **String voltage says whether it is day**, and it survives our own
+        commands: a forcible discharge holds the tracker down and the panels
+        sit at open-circuit voltage — 374 V at 0.00 A — so voltage still reads
+        daylight. What it cannot do is tell a bright morning from a dim one. An
+        overcast dawn stands near 300 V while the array makes 50 W, and
+        refusing to command there costs the house a discharge it needs, for
+        nothing. Measured across three dawns on the reference installation,
+        that band runs 45-50 minutes.
+
+        **Harvested power says how much there is to lose**, and is the quantity
+        that actually matters — but a low reading is only honest while this
+        driver is released. Once a command stands, the production it suppresses
+        is our own artefact, and gating on it would read a curtailed roof as
+        darkness, keep commanding, and keep the roof curtailed. That is the
+        failure the voltage test was introduced to stop: a discharge commanded
+        at 04:32 against a dark sky, still standing at 09:22.
+
+        A *high* reading is a different matter. Our command can only ever push
+        this number down, never up, so production above the threshold is
+        decisive whatever the driver is doing — that is what hands control back
+        at sunrise without waiting for anything.
+
+        A low reading while commanding is the one ambiguous case, and the only
+        one that needs a probe: let go, wait for the tracker to find its own
+        maximum, and read again. A probe costs seconds of curtailment. Not
+        probing costs a day of it.
+        """
+        now = asyncio.get_running_loop().time()
+
+        voltages = [
+            snapshot.get(f"pv{index}_voltage")
+            for index in range(1, self._pv_strings + 1)
+        ]
+        if any(volts is not None for volts in voltages):
+            self._strings_lit = any(
+                volts is not None and volts > _PV_LIT_VOLTAGE_V for volts in voltages
+            )
+
+        if not self._strings_lit:
+            # Night needs no second opinion, and nothing is being curtailed.
+            self._pv_lit = False
+            self._pv_probe_until = 0.0
+            self._pv_verdict_monotonic = now
+            return
+
+        harvest = snapshot.get("solar_power")
+        if harvest is None:
+            # No reading is not evidence either way; the last verdict stands.
+            return
+
+        if float(harvest) > _PV_HARVEST_THRESHOLD_W:
+            self._pv_lit = True
+            self._pv_verdict_monotonic = now
+            self._pv_probe_until = 0.0
+            return
+
+        released = (
+            self._last_written_w in (None, 0)
+            and now - self._last_write_monotonic >= _PV_PROBE_SETTLE_S
+        )
+        if released:
+            # Nothing of ours is holding the tracker down, so a low reading is
+            # the roof's own answer and the battery is ours to command.
+            self._pv_lit = False
+            self._pv_verdict_monotonic = now
+            self._pv_probe_until = 0.0
+            return
+
+        if self._pv_probe_until:
+            if now >= self._pv_probe_until:
+                # The release never reached the hardware — the write was
+                # refused, or something else is commanding this battery.
+                # Stop pretending to measure and wait out the interval.
+                _LOGGER.debug(
+                    "Huawei driver: PV probe expired without a usable reading"
+                )
+                self._pv_probe_until = 0.0
+                self._pv_verdict_monotonic = now
+            return
+
+        if now - self._pv_verdict_monotonic >= _PV_PROBE_INTERVAL_S:
+            _LOGGER.info(
+                "Huawei driver: releasing briefly to re-read the roof — %.0fW "
+                "reported while commanding is not evidence of darkness",
+                float(harvest),
+            )
+            self._pv_probe_until = now + _PV_PROBE_WINDOW_S
 
     def _should_write(self, applied: int) -> bool:
         """Whether this set-point is worth four Modbus writes and a 10 s ramp."""
--- a/custom_components/omnibattery/drivers/huawei.py
+++ b/custom_components/omnibattery/drivers/huawei.py
@@ -1125,7 +1125,21 @@ class HuaweiSolarDriver(BatteryDriver):
         The subtraction deliberately excludes this battery's own contribution:
         the ceiling has to describe what PV occupies, not what the battery is
         currently doing, or the limit would chase its own output and oscillate.
+
+        Zero while the PV gate is closed, which is the honest answer and not a
+        special case. The allocator hands out shares against this limit and
+        then believes the battery was served; leaving the static envelope in
+        place would have it allocate a discharge that ``apply_setpoint`` is
+        about to refuse, so the fleet under-delivers and no other battery picks
+        up the slack. Saying zero up front moves that share to a battery that
+        can actually deliver it.
+
+        There is no charge equivalent — the allocator has no comparable seam on
+        that side, so a refused charge is still invisible to it. See §13.9 of
+        the driver assessment.
         """
+        if self._hands_off():
+            return 0
         ceiling = data.get("inverter_max_power")
         ac_power = data.get("inverter_ac_power")
         battery_power = data.get("battery_power")
--- a/custom_components/omnibattery/drivers/huawei.py
+++ b/custom_components/omnibattery/drivers/huawei.py
@@ -459,6 +459,9 @@ class HuaweiSolarDriver(BatteryDriver):
         self._pv_probe_until = 0.0
         # When the verdict last came from a reading this driver could trust.
         self._pv_verdict_monotonic = 0.0
+        # Last harvest this driver had reason to believe, in watts. None when
+        # the standing command makes the reading its own artefact.
+        self._pv_harvest_w: Optional[float] = None
         self._serial: Optional[str] = None
         # Last command actually written, so the deadband can compare against what
         # the hardware was told rather than against what it currently delivers.
@@ -795,7 +798,7 @@ class HuaweiSolarDriver(BatteryDriver):
             min(self._ceiling("charge"), int(net_power_w)),
         )
 
-        if applied != 0 and self._hands_off():
+        if self._refuses(applied):
             # A forcible command on this hybrid is not a request but a ceiling:
             # the inverter produces exactly what the command asks for and
             # curtails the rest of the roof. Measured with a 315 W charge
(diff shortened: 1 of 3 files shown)


ffunes:
Reopening to pick up the fixed base — see the comment that follows.

ffunes:
Retargeted to `release/v1.5.0` (1.4.0 is closing out its betas), and reopened to re-run the checks against the fixed base — details in #380.

On the gate: you're right and I was wrong. The table settles it. A discharge standing through sunrise holds the tracker down, so 374 V at 0.00 A and a 300 V / 50 W overcast dawn are the same reading to a harvested-power threshold, and gating on it brings back the 04:32→09:22 command. Voltage is the only signal that survives our own curtailment, which is the part I hadn't thought through.

Build the version you described:

- read `solar_power` (32064) only while the driver is released, where it is the true harvest;
- latch the verdict while a command stands, since the reading is then our own artefact;
- release periodically to re-sample.

Choose the threshold and the re-sample period from what you can measure — yours is the only installation that can. Put them in the driver assessment with the dawns behind them and I'll take the numbers as given. If a dawn shows the periodic release is itself disruptive, say so and we ship the plain voltage gate with the limitation documented; a Huawei battery being unavailable for an hour at dawn is a smaller failure than one discharging all day.

`dc_coupled` staying out with the rest of that work was the right call.

sphings79:
Built, and rebased onto `release/v1.5.0` — which is green now, so the suite is 1949 passed, 10 skipped, no failures at all.

### The asymmetry that makes harvested power usable

Working through it, the reading is not uniformly untrustworthy while a command stands — it is untrustworthy in exactly one direction:

- A **low** reading proves nothing, because the production it reports is what our own command left of the roof. Gating on it reads a curtailed roof as darkness, keeps commanding, keeps it curtailed.
- A **high** reading is decisive whatever the driver is doing, because a forcible command can only ever push that number down, never up.

So the sunrise hand-back needs no probe at all: production above the threshold releases the battery on the next telemetry poll. The probe is only for the one genuinely ambiguous case — commanding, strings lit, roof quiet — which is the dim-dawn band and nothing else. That turned out much narrower than I described it in my last comment.

### The numbers, and where they came from

Three dawns off the reference installation, first non-zero watt to a sustained 150 W:

| Dawn | First harvest | Sustained >150 W | Band |
|---|---|---|---|
| 30.08 | 06:55 | 07:45 | 50 min |
| 31.08 | 06:15 | 07:00 | 45 min |
| 01.09 | 06:48 | 07:38 | 50 min |

`_PV_HARVEST_THRESHOLD_W = 150.0`, deliberately low — too high curtails a roof that is genuinely producing, too low only costs availability in a band where there is nothing to curtail. `_PV_PROBE_INTERVAL_S = 120.0`, settle 15 s against a tracker that recovered in 6 s when measured.

Both are first values, not final ones. String voltage and lux were excluded from the recorder on that installation — I found this while pulling these dawns — so the 45-50 minutes is a **lower bound**: it measures from the first watt, not from when the strings cross 100 V, which is earlier. Recording is in place since yesterday evening, so the next dawns will give the real figure and I will correct the constants and the table against it.

### One thing the tests turned up

The release is throttled like any other write. A verdict change does not jump `_MIN_WRITE_INTERVAL_S`, so hand-back and take-back each wait out the ramp interval, and `apply_setpoint` honestly reports the command still in force meanwhile. A probe therefore costs roughly 20 s in every 120 s of the dim band rather than the 15 s settle alone.

I left it that way — the throttle exists because rewriting mid-ramp makes this inverter derate the array, which is the same fault this whole gate is about. But it is the real cost, it is pinned in `test_the_release_waits_out_the_ramp_throttle`, and if 17 % looks too rich the interval is the knob.

### Tests

Eleven new ones: the overcast dawn commanding normally, the threshold sitting on harvest rather than voltage, production believed at once while commanding, a quiet roof under command **not** taken as darkness, the probe finding sun and finding nothing, a probe whose release never reached the hardware, darkness needing no probe, an unreadable roof leaving the verdict alone, and the throttle above.

§13.8 of the driver assessment carries the dawn table and the asymmetry.

### Still open from your side

Your fallback offer stands unused for now — if a dawn shows the periodic release is itself disruptive, I will say so and we ship the plain voltage gate with the limitation documented. I would rather have that answer from a measured dawn than from my own reasoning about one.

ffunes (commented):
Verified in a clean worktree on `release/v1.5.0`: **1949 passed, 10 skipped, 0 failures**. The `test_charge_delay.py::test_balance_needs_charge_unlocks_low_forecast` failure you flagged does not occur against this base — it is already fixed here, so that note can go.

The gate design is right, and the asymmetry it rests on is the key observation: a *high* harvest reading is decisive whatever the driver is doing, because our command can only ever push that number down; a *low* one only counts while we are released. That is the correct answer to the question I asked, and better than the plain harvest gate I was reaching for.

One detail worth calling out because it is easy to get wrong and you got it right: `released` is checked *before* the probe-expiry branch, so a release that lands late but does land still yields a valid verdict instead of being abandoned as "never reached the hardware".

## Blocking: the gate's inputs are not declared as control dependencies

`control_dependency_keys` does not include `solar_power` or `pv{n}_voltage`. The coordinator stops polling any key whose entity is disabled and which is not a declared dependency (`coordinator.py`, the `is_disabled` / `is_dependency` branch).

So a user who disables the **Solar Power** entity silently disables the whole gate: `harvest is None` → early return → `_pv_lit` stays at its initial `False` forever → the driver commands in full sun. That is the 04:32→09:22 failure reintroduced by an entity toggle.

Reproduced against your branch, both pass:

```python
@pytest.mark.asyncio
async def test_only_solar_power_disabled():
    driver = _driver(_fake_client(_lit_blocks()), hass=_hass_with_services())
    # what the coordinator asks for once Solar Power is disabled
    await driver.read_telemetry(["battery_power", "mppt1_power", "mppt2_power",
                                 "force_mode", "set_charge_power", "set_discharge_power"])
    assert driver._strings_lit is True    # it knows it is daylight
    assert driver._pv_lit is False        # ...but never closes the gate
    assert (await driver.apply_setpoint(-2000, read_back=False)).net_power_w == -2000
```

The worse variant — MPPT Power and PV Voltage disabled too, so `_strings_lit` freezes at `False` and it commands without even knowing it is day — passes as well. `pv{n}_voltage` is `enabled_by_default: False` and only arrives dragged along by `mppt{n}_power` via `_DERIVED_FROM`, so there are two user-facing toggles between the gate and its evidence.

One-line fix: add `solar_power` and the `pv{n}_voltage` keys to the frozenset, with a comment saying why — the same shape as the `dynamic_discharge_limit_w` inputs block right above it. Plus a test that polls without them.

## The fleet is not told the battery declined

When the gate closes, `apply_setpoint` returns `ok=True, net_power_w=0` and the coordinator merges `force_mode: stop` into `coordinator.data`. Coherent for the PD loop — but the allocator has already given this battery its share and believes it was served, so the fleet under-delivers and no other battery picks up the slack.

There is already a seam for this that needs no control-layer code: `dynamic_discharge_limit_w` narrows what the allocator sees, and `_apply_driver_dynamic_limit` consumes it. `return 0` while `_hands_off()` closes the discharge case entirely. There is no charge equivalent — worth stating that explicitly in §13.9 rather than leaving it implicit.

## The decision I actually owe you: charge

The block is total, and at a 150 W threshold that is all daylight. For a dynamic-pricing user it means **a cheap midday slot is impossible to charge into, permanently**. §13.8 states the cost honestly, so it is disclosed — but it is the largest functional loss in the PR and deserves a decision rather than a paragraph.

And your own measurement suggests the charge case may be wider than it needs to be. 315 W commanded, 288 W harvested: if a forcible charge command *is* the production ceiling, then a command **above** the current harvest costs nothing at all. Gating charge on `applied < harvest` rather than blanket would keep cheap-hours grid charging alive on a lit day.

I will not assert that from one data point. But it is the measurement I would want before merge, and you have the hardware. Discharge stays blanket either way — it holds the tracker down completely (374 V at 0.00 A), there is no nuance there.

## Minor

- **Single-installation thresholds.** `_PV_LIT_VOLTAGE_V = 100.0` and `_PV_HARVEST_THRESHOLD_W = 150.0` are measured on one ~374 V array. A small string can genuinely produce below 100 V. I am not asking for config options now — just that the comment says these came from one installation, so the second user who disagrees does not cost a day of diagnosis.
- **Probe duty cycle in the dawn band.** ~20 s refused in every 120 s, during exactly the 45-50 minutes the harvest gate was added to recover — about 17% of that window. The tests document it honestly; just confirming you accept it.
- **`_pv_lit` no longer means "lit"**, it means "hands off"; `_strings_lit` is now the literal one. A rename would save the next reader a pass. Optional.

## Docs

§13.8-13.10 and the EMMA register map are good. §13.9 — the second battery reading as household load to the hybrid's energy manager, with 1391 W of PV over a 529 W house, one battery taking 1110 W while the other gave up 205 W and the meter at 3 W — is the most valuable finding in the PR, and it is exactly what the #335 control layer needs to know.

---

**Verdict:** approve once `control_dependency_keys` is fixed. Everything else is conversation, not a blocker.

sphings79:
Thanks — the blocker was real, and your reproduction passed against the branch before I changed anything, which is the part that matters.

## Fixed: the gate's inputs are declared

`solar_power` and `pv1..4_voltage` are in `control_dependency_keys` now, with a comment saying why: without them `_update_pv_gate` returns early, the verdict never leaves its initial `False`, and one toggle in the entity list puts the 04:32→09:22 failure back in full sun with no warning. Your point about there being *two* switches — the string voltages being `enabled_by_default: False` and only arriving dragged along by `mppt{n}_power` — is in the comment too.

Three regression tests: the keys are declared; the gate still closes when polling only the declared set; and a dim dawn still reads as commandable through the same path.

## Also fixed: the fleet is told

You were right that this needs no control-layer code, and right about which seam. `dynamic_discharge_limit_w` returns 0 while `_hands_off()`, so the share moves to a battery that can deliver it instead of being allocated here and refused. Four tests, including the probe window — while probing the driver has let go, so it cannot serve a share either.

The missing charge equivalent is now stated outright in §13.9 rather than left implicit, with the note that on a DC-coupled hybrid the loss is smaller than it sounds because the inverter is charging from the strings anyway — but that it is real and belongs to the control layer.

## Thresholds

Comment now records that both come from one ~374 V array on a SUN2000-8K, and that a second installation disagreeing is a reason to look at the numbers before the logic.

## The charge measurement — running tomorrow

Agreed this is the right question, and it is the one thing here I will not answer from reasoning. Your reading of the 315 W / 288 W data point is the hypothesis: if a forcible command *is* the production ceiling, a command above the current harvest costs nothing, and `applied < harvest` keeps cheap-hours charging alive on a lit day.

I could not run it tonight — no sun, strings at 0.0 V, roof at 0 W, nothing to curtail. Tomorrow in daylight I will command a charge well above the standing harvest and watch whether DC production drops or holds, across several harvest levels rather than one, since the interesting case is precisely where the command crosses the harvest. I will report the numbers whichever way they fall — including if they say your hypothesis is wrong and blanket is the honest answer.

Two things I would rather you decide than assume:

- **Does that measurement block the merge?** Your verdict says approve once `control_dependency_keys` is fixed and everything else is conversation, but the charge paragraph reads more like a precondition. I am happy either way — merge now and I will open the charge refinement as its own PR against the measurement, or hold #381 until tomorrow evening.
- **The discharge side stays blanket regardless**, as you said. 374 V at 0.00 A leaves nothing to weigh.

## Declined, with reasons

**The `_pv_lit` rename.** You are right that it means "hands off" and that `_strings_lit` is the literal one. I am leaving it for now: renaming a field across ~20 test assertions mid-review makes your re-read harder for a readability gain, and the diff is already carrying two fixes from this round. Happy to do it as a follow-up once this lands, or now if you would rather read it renamed.

**The probe duty cycle**, confirmed accepted: ~20 s refused in every 120 s, about 17 % of the 45-50 minute dawn band. It buys the other 83 % of a window that was previously refused entirely, and `_PV_PROBE_INTERVAL_S` is one constant if a dawn says otherwise. Tomorrow's dawn is the first with string voltage in the recorder, so I will have the real band figure — measured from the strings crossing 100 V rather than from the first watt — and will correct both the constant and the §13.8 table against it.

Suite: 1956 passed, 10 skipped, 0 failures on `release/v1.5.0`.

sphings79:
First dawn with string voltage in the recorder, so the band is measured rather than bounded — and it is roughly **double** what I told you.

## The measured band: 97 minutes

| Dawn | Gate closes | Sustained >150 W | Band | Measured from |
|---|---|---|---|---|
| 30.08 | 06:55 | 07:45 | 50 min | first watt (lower bound) |
| 31.08 | 06:15 | 07:00 | 45 min | first watt (lower bound) |
| 01.09 | 06:48 | 07:38 | 50 min | first watt (lower bound) |
| **02.09** | **06:12:56** | **07:50** | **97 min** | **100 V on the strings** |

The strings went from 0 V to 123.8 V in a single 30-second poll at 06:12:56 — the inverter waking, not a gradual rise — which is **23 minutes before sunrise** at 06:36. The array made 1 W or less until 07:12 and did not hold above 150 W until 07:50. The 07:50 bucket is the first whose *minimum* clears the threshold, not just its mean.

The earlier three are now labelled as lower bounds in §13.8 rather than quietly compared against the new one.

## Why I did not just report that number

A forcible discharge was standing for part of the window, which suppresses production by itself — so the harvest curve alone could not tell "dark" from "curtailed by us", which is the same ambiguity this whole gate is about. The irradiance sensor is independent of the inverter and settles it:

- 5 lx at 06:21, 98 lx at 06:42, 250 lx at 07:10

There was nothing to curtail. The same sensor shows the morning was cloudless — 1486 lx climbing smoothly to 1955 lx across the crossing, no dips — so **97 minutes is the clear-sky figure and an overcast dawn is longer**, not shorter.

Incidental, and not the gate's doing: the discharge that was standing came from the inverter's own self-consumption after our command released at 06:24:12, and ran until the battery hit its 5 % floor at 06:41.

## What the measurement changed: the threshold needs a band

The value stays at 150 W — it separated dark from producing cleanly. But the ramp **crawls across it rather than stepping over**:

| Bucket | mean | min | max |
|---|---|---|---|
| 07:40 | 137 | 123 | 156 |
| 07:45 | 100 | 0 | 176 |
| 07:50 | 168 | 152 | 214 |

Twenty minutes wobbling either side of 150 W, with irradiance rising smoothly throughout — so the wobble is the array, not the weather. A bare comparison flips the gate on every one of those, and each flip costs a write that this inverter answers by derating the array. That is the fault the gate exists to avoid, reintroduced by the gate itself.

`_PV_HARVEST_HYSTERESIS_W = 50.0`, latching, same shape as `SURPLUS_GUARD_HYSTERESIS_W`: the band above on the way in, below on the way out. Pinned in `test_the_dawn_ramp_does_not_flip_the_gate_on_every_wobble`, which walks the measured sequence 123 → 176 → 214 → 152 → 123 W and asserts a single transition rather than four.

I would not have found this from the earlier dawns, because without voltage recorded I was reading the curve from the first watt and never looked closely at the crossing.

Suite: 1957 passed, 10 skipped, 0 failures.

## Still to come today

The charge measurement, mid-morning — whether a command *above* the current harvest costs anything. That one needs stable irradiance, so it waits for the sky to settle rather than running at a fixed time.

sphings79:
A question rather than a finding, because I cannot reproduce it on my installation and would rather ask than assert.

## Does a self-curtailing inverter read as darkness?

Checking whether last night's household load had skewed the dawn measurement, I ended up establishing the opposite: consumption ran 670–965 W all night with the battery at its 5 % floor, so there was ample demand, and the roof still reported 0–1 W. That makes the darkness conclusive — with that much demand, any available watt would have been harvested.

Which is the point. **The harvest reading only means "what the roof can make" while something wants the power.**

On a hybrid running self-consumption, a full battery and a quiet house mean the inverter curtails on its own — production tracks load, not irradiance. If that load sits below the threshold, `solar_power` reads under 150 W in full sun, the gate opens, and the driver commands into a lit array.

That is the same trap twice removed. First it was our own forcible command suppressing the evidence. Then it was the plain harvest gate you proposed, which the asymmetry fixed. This would be the *inverter's* own regulation suppressing it — and the asymmetry does not help, because a low reading is still a low reading.

**Nor does the probe.** Releasing tells us nothing here: the inverter goes on curtailing whether we command or not, so the re-sample returns the same suppressed number. Every discriminator this driver has is on the DC side, and on the DC side the two states are identical.

## Why I cannot settle it here

The reference installation exports — 12.10 kWh on 27 August — so its inverter runs at MPP regardless of what the house is doing, and `solar_power` is honest all day. I can't produce the failing condition without changing the export configuration, which I am not going to do on a live installation to chase a hypothetical.

It seems reachable on an installation with zero-export or a hard export limit, a battery that fills by midday, and a household whose base load is under 150 W. Whether that combination exists among your users I have no idea.

## What I'd want to know

1. **Does a SUN2000 under a zero-export limit actually curtail DC production**, or does it produce at MPP and dump the difference? If it is the latter, this question dissolves and you can ignore the rest.
2. If it does curtail — is a battery near `charging_cutoff_capacity` a good enough discriminator? A low harvest is not evidence of darkness when the battery cannot absorb anything, so the gate could fall back to voltage-only there. It costs nothing on the charge side, since a full battery has nothing to gain. It does cost the discharge side: a full battery at an overcast dawn would be refused again, which is the hole you asked me to close in the first place. Rare, but the same shape.
3. Or is this thin enough to document as a limitation and leave alone? I would not object — every version of this gate trades one misread for another, and the two I have measured are both worse than this one.

Not blocking anything from my side. The charge measurement is still to come today and I will report it separately.

sphings79:
Withdrawing my first question — I asked the one person here who has no Huawei to answer a Huawei hardware question. Mine to answer, and it turns out it needs no hardware at all.

**A SUN2000 under an export cap must curtail, by conservation.** PV has exactly three destinations: house load, battery, grid. With the battery full and export capped, the only remaining sink is the house, so DC production has to equal house load. There is no dump load in a string inverter — curtailment *is* how any PV inverter limits output, by walking the tracker off the maximum power point. This is not Huawei-specific and does not need testing.

So the scenario is not speculative, it is forced: wherever export limitation is configured and the battery fills before the day is over, `solar_power` reports house load rather than irradiance for the rest of that day. If that load sits under the threshold, the gate opens in full sun.

That said, it does not follow that anyone hits it. It needs an export limit, a battery that fills, *and* a base load under 150 W at the same time — and the third is the demanding one for a house with a hybrid and a battery in it.

What is genuinely yours to decide is what remains from that comment:

- Is a battery near `charging_cutoff_capacity` a good enough discriminator, accepting that it costs the discharge side at a dawn with a full battery — the hole you asked me to close, back in a narrower form?
- Or is this thin enough to document as a limitation and leave alone?

I lean toward documenting it. Every version of this gate trades one misread for another, and the two I have actually measured are both worse than this one. But it belongs in §13.8 as a stated limit rather than as something nobody noticed, and I would rather have your call than assume it.

Sorry for the detour — the physics was available to me the whole time and I put it in front of you as a question instead.

sphings79:
Measured. **Your hypothesis holds, cleanly, and the charge side should not be blanket-blocked.**

## The numbers

Battery at 5 % with room to take anything, cloudless, irradiance rising smoothly throughout. Commands sent straight through `huawei_solar.forcible_charge` with `duration: 2` so nothing could latch.

| Time | State | Roof DC | Lux | Balcony | Roof ÷ Balcony |
|---|---|---|---|---|---|
| 10:20:13 | uncommanded | 1839 W | 15809 | 394 W | 4.66 |
| 10:20:38 | **charge 400 W** | **363 W** | 15707 | 401 W | **0.91** |
| 10:20:49 | charge 400 W | 363 W | 15707 | 403 W | 0.90 |
| 10:21:12 | charge 400 W | 363 W | 15707 | 403 W | 0.90 |
| 10:21:19 | released, +7 s | **1936 W** | 15733 | 403 W | **4.80** |
| 10:21:53 | **charge 3000 W** | 2020 W | 16278 | 407 W | 4.96 |
| 10:22:14 | released | 2031 W | 16516 | 407 W | 4.99 |

**A command below the harvest is the production ceiling, exactly.** 400 W commanded, 363 W produced — the array fell to a fifth of what it had been making one sample earlier, and stayed there across three readings.

**A command above the harvest costs nothing.** 3000 W against roughly 1900 W of capability: production carried on tracking irradiance, ratio 4.96 against 4.66 and 4.80 uncommanded.

**Recovery is 7 seconds**, which corroborates the 6 s in the original observation and means the probe's 15 s settle has margin.

The balcony array is on a Shelly plug with no electrical relationship to the hybrid, and the irradiance sensor none either. Both moved by about 1 % while the roof moved by 5×. This is not weather.

## What it means for the gate

Blanket-blocking charge gives away real energy for no reason. The rule you proposed is the right one: **refuse a charge only when `applied < harvest`.**

It is also self-stabilising, which I did not expect until I looked at it:

- While an *allowed* command stands (at or above harvest), the cap is not binding, so `solar_power` keeps reporting true MPP — the reading stays trustworthy exactly in the case where we are commanding.
- A refused command is never sent, so it never suppresses the evidence that refused it.

One edge worth handling: the sun can rise past a standing command. Allow 2000 W at 1900 W of harvest, the day brightens to 2500 W, and the command is now a cap. It is detectable without extra telemetry — `harvest` converges on `applied` — and the answer is to raise the command or let go.

## What I would build, if you want it

- Split `_hands_off()` by direction. Discharge stays blanket: 374 V at 0.00 A leaves nothing to weigh.
- Charge is refused only when `applied < harvest`, using the same latched harvest the gate already keeps, so a suppressed reading can never authorise a command.
- Re-check a standing charge against harvest each cycle, and release or raise when the two converge.
- Tests off these numbers: 400 W against 1839 W refused, 3000 W against 1936 W allowed, and the convergence case.

Say the word and it goes on this branch. I have not built it yet because it changes what the gate is for, and that is yours.

## Two caveats

**`sensor.batterien_erzwungene_ladung` is not a readback.** It has read `Stopped` since 06:24 this morning and never moved through any of this, including while a command was demonstrably capping the array at 363 W. I verified cleanup functionally instead — production back to ratio 5.28, above baseline — and by the 2-minute `duration` expiring on its own. Worth knowing if anything ever leans on that entity.

**The commanded power is not what the battery takes.** 3000 W commanded put 849 W into the battery; 3500 W earlier put in 1536 W briefly, then bounced. The command's grip on *production* is exact; its grip on *battery flow* is not, and the difference goes to house load and the grid. That does not affect the result — the ceiling is on production — but it means the command is not a battery set-point in the way it looks.

ffunes:
Verified the merge against current `release/v1.5.0` locally: no conflicts, and the full suite is 2074 passed, 10 skipped, 0 failures. `test_balance_needs_charge_unlocks_low_forecast` no longer fails — it came in with the old base and went out with it.

So the mechanics are fine. What I want before merging is the thing your own last measurement asks for.

**Build the direction split.** 400 W against 1839 W producing 363 W, and 3000 W against 1936 W producing 2020 W, is as clean as a result gets. `_hands_off()` is symmetric today, so every daylight charge is refused — and on this integration that is precisely the cheap dynamic-pricing slot, which is the one charge the user went looking for. Blanket on discharge, since 374 V at 0.00 A leaves nothing to weigh; `applied < harvest` on charge, off the latched harvest; plus the convergence re-check for a command the sun rises past. Your four bullets as written.

The self-stabilising property you found is what convinces me this is a better gate rather than merely a looser one: an allowed command is not a binding cap, so the reading stays honest exactly while we are commanding, and a refused command is never sent, so it cannot suppress the evidence that refused it.

**Export cap: document it, don't discriminate on SOC.** The conservation argument is right and needed no hardware. But falling back to voltage-only near `charging_cutoff_capacity` buys back the dawn hole with a full battery — the same shape you were asked to close — and the failing combination needs an export limit, a battery that fills, *and* a base load under 150 W at the same time. The third is the demanding one in a house with a hybrid in it. §13.8 as a stated limit, which was your lean; consider it confirmed.

**Rebase while you are in there.** The branch is seven commits behind, and one of them is the mixed-fleet control layer (the other half of #335). Your `dynamic_discharge_limit_w` docstring argues about an allocator that has changed underneath it — the reasoning still holds as far as I can see, and the merge is green, but nothing exercises a hands-off Huawei inside a mixed fleet. One test for that, please.

Nothing else from me. The EMMA map and §13.8–13.10 go in as they are. And `sensor.batterien_erzwungene_ladung` costs us nothing: the driver already declares `setpoint_confirm_reliable=False`, so no code leans on that entity.

sphings79:
All four done, rebased onto current `release/v1.5.0` — 2081 passed, 10 skipped, 0 failures.

## The direction split

`_hands_off()` is gone; `_refuses(applied)` decides per command. Discharge outright, since it holds the tracker down whatever its size. Charge only when `applied < harvest`, against the latched harvest.

## The convergence case turned out not to need its own logic

I had it down as a fourth mechanism and it collapsed into the third. What decides is whether the *reading* can be believed:

- nothing standing → it is the roof's own answer;
- a charge clear of the harvest by the hysteresis band → the ceiling is never reached, so the tracker is still at its own maximum and the reading is true MPP;
- a charge the harvest has converged on, or any discharge → our own artefact, so `_pv_harvest_w` is `None`.

`None` refuses the charge, the release restores a true reading, and the decision is retaken against it. The sun rising past a standing command corrects itself without a special case, and an allowed charge keeps its own evidence alive in the meantime — which is the property that makes the whole thing safe.

**One ordering bug, caught by the test for exactly that case.** The decisive-high branch was latching the harvest before the trust check ran. A high reading is a valid floor for the *verdict*, since a command can only push the number down — but a floor is not a maximum, and only a maximum is safe to compare a charge against. Under a binding command you read back the command itself and would have believed it. The trust check now runs first.

## The mixed fleet

Two tests through `_apply_driver_dynamic_limit`, which is the seam the allocator actually reads: a refused hybrid reports 0 and yields its share while the AC battery beside it keeps its full envelope, and after dark it takes the envelope back. You were right that nothing exercised it — the existing tests called `dynamic_discharge_limit_w` directly and never saw it beside another battery.

## Export cap

Documented, not guarded, as you decided. §13.8 now carries it as a stated limit with the reasoning: with the battery full and export capped, production must equal house load, so a quiet house reads as darkness; neither signal separates that from real darkness, and releasing changes nothing because the inverter curtails whether commanded or not. Plus why SOC is the wrong discriminator — it buys back the dawn refusal for a full battery, which is the larger fault.

§13.8 also gains the measurement table, and the decision report no longer says "control after dark only": it now separates discharge from charge.

## On the rebase

The only conflict was `mkdocs.yml` — ac82c3b had already added the EMMA nav line my branch was adding, and dropping yours would have unlinked the field-evidence page. Resolved to your version.

`dynamic_discharge_limit_w`'s docstring still argues about the allocator, and I re-read it against `control/` rather than assuming: the claim is that the allocator hands out shares against this limit *before* deciding, so a battery about to refuse must say so here. That still holds — it is `_apply_driver_dynamic_limit` in the control path, unchanged in 216bbed. I left the wording alone.

## One thing left over from #382

`primary-battery-field-evidence.md` line 54 still says `_scarce_solar_day`, the fork's name. In `control/charge_order.py` it is `scarce_solar_day`. A dead reference rather than a wrong claim, but it sends the next reader looking for a function that is not there. Happy to fix it here or leave it to you.

Issue #350, opened 2026-08-27
Multi-pack Venus D: one finished pack arms the BMS-cutoff detector for the whole battery


sphings79:
## Environment

- Marstek Venus D, **6 battery packs**, 15.36 kWh, 2500 W
- Running a fork of `release/v1.4.0`. `control/weekly_full_charge.py` is unmodified there, and the fork's `__init__.py` diff contains no line touching `hysteresis` or `bms_cutoff` — so what follows is upstream behaviour.

## Summary

On a Venus D the packs are topped up **one after another** during the final absorption phase. As soon as the first pack completes, its cells sit above `NORMAL_BALANCE_TAPER_CELL_VOLTAGE`. Because `max_cell_voltage` is a single value for the whole battery, `tick_bms_cutoff()` stays armed for the rest of the charge while four or five packs are still at 90 %.

A pack handover then produces a brief moment of ~0 W draw while a charge is still commanded, which is counted as a BMS refusal. `is_battery_full()` turns true and the charge hysteresis latches at the **aggregate** SoC — 93 % on this installation, with `charge_resume_threshold` 91 %.

**This is self-correcting and the cost is small.** The staleness check

```python
if (coordinator._hysteresis_base_soc is not None
        and coordinator.max_soc > coordinator._hysteresis_base_soc):
```

cleared the latch after 12.5 minutes and charging resumed normally, without the SoC ever dropping. Since the battery only accepts ~200 W in absorption, that pause cost about 0.04 kWh. Repeated once per pack it would be roughly 0.25 kWh per full charge.

So this is a false positive in the detector rather than a yield problem. Reporting it because the detection itself looks unintended on multi-pack hardware, not because it hurt much here.

## Measured (27 Aug 2026)

| Time | Observation |
|---|---|
| 13:56 | **Pack 1 reaches 100 %.** Its max cell voltage settles at 3.481 V |
| 13:58–15:17 | Pack 2 tops up at a flat 140–145 W. Packs 3–6 unchanged at 90.0 % |
| 15:17:37 | **Pack 2 reaches 100 %.** `set_charge_power` → 0, `force_mode` → `None` |
| 15:17:40 | `charge_hysteresis_active: on`, `base_soc: 93`, `charge_resume_threshold: 91` |
| 15:30:11 | Latch clears on its own, `set_charge_power` → 200. SoC had *risen*, not fallen |
| 16:36 | SoC 95 %, charging normally, hysteresis re-armed at `base_soc: 100` |

[...]

PR #382, opened 2026-09-01
Field evidence: the charge order on a mixed AC/DC installation

sphings79:
The field evidence from #335, as you asked, against `release/v1.5.0`.

### What changed from the version on that branch

The implementation claims are now observations of the build that produced them. Concretely:

- A preamble says outright that **none of the control-layer behaviour described is part of Omnibattery** — the charge order, the scarcity threshold and the primary-battery selection were prototyped on a fork and are being reimplemented in `control/`. Where a rule is named, it is the rule the build under test applied.
- The same preamble draws the line the other way: the *hardware* findings do hold generally. A hybrid's response to a forcible command, and what a second battery looks like to an energy manager, are properties of the equipment rather than of any control layer.
- "Before the fix" / "after the fix" are now "under the fleet-room threshold" / "under the DC-room threshold", which is what actually distinguishes the two days.
- "Scarcity is now measured against…" is past tense and attributed to the build.
- The daylight-command guard is attributed to #381 rather than described as existing behaviour.
- "The defect" is now "The rule under test, and how it failed", with a line saying why it is worth carrying forward: the failure is in the question, not in the code that answers it.

### What is in it

- **The charge order across two days**, with the controlled comparison at matched irradiance rather than daily totals: 26.08 10:00 at 2101 W of PV, the AC battery at −12 W on its 12 % floor while the two-thirds-full hybrid took the entire surplus; 27.08 07:00 at 2354 W, the AC battery at +964 W while the hybrid — *lower*, at 6.5 % — took +766 W.
- **AC/DC conversion measured at both ends**: 6.5 % at full load, 27 % at absorption power.
- **The absorption tail** on sequential-pack hardware, and the ~200 W device-reported ceiling above 90 %.
- **Telling shade from control** via a separate irradiance instrument, with the honest note that it was added mid-day and has no recorded history before 09:37.
- **The false BMS cutoff** (#350), now with the 1 September re-measurement against your fix: two pack handovers passed with the detector armed, at the same 93 % aggregate where 27 August stalled, and neither latched.
- **Limitations**, unchanged and deliberately unflattering — two days, one brand pairing, the sunnier day is the one after the change, the AC battery never reached 100 %, and the handover between two batteries that both have room has still not been observed.

Added to the mkdocs nav next to the Huawei driver assessment.

### One gap I should name

The irradiance sensor was excluded from the recorder on that installation, so there is no lux history for either day and none is quoted — PV power stands in for irradiance throughout. The same turned out to be true of the inverter's per-string voltage and current, which I found while starting the dawn measurements for #381.

Both are now recorded, as of 1 September 16:53: a `recorder: include: entities:` block overriding the `sensor.*_spannung` / `sensor.*_strom` / `sensor.*_illuminance*` globs, with long-term statistics suppressed via `customize_glob` and `state_class: ""` so only the 10-day history is kept. Verified live — the states arrive, the statistics stay empty.

That fixes nothing retroactively: 26 and 27 August have no string or lux history and never will, so the limitation stands as written in the document. It does mean the dawn measurements for #381, and anything measured on this installation from here on, will have irradiance and per-string data alongside the power figures.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -183,4 +183,5 @@ nav:
       - Architecture: reference/architecture.md
       - Battery driver requirements: reference/driver-requirements-template.md
       - Driver assessment · Huawei: reference/driver-assessment-huawei.md
+      - Field evidence · mixed AC/DC charge order: reference/primary-battery-field-evidence.md
   - Changelog: changelog.md
--- a/site-docs/reference/primary-battery-field-evidence.md
+++ b/site-docs/reference/primary-battery-field-evidence.md
@@ -1,0 +1,308 @@
+# Field evidence: the charge order on a mixed AC/DC installation
+
+Measurements from a mixed AC/DC installation, recorded on 26 and 27 August
+2026 either side of a change made to the experimental build running on it.
+
+**None of the control-layer behaviour described here is part of Omnibattery.**
+The charge order, the scarcity threshold and the primary-battery selection were
+prototyped on a fork (#335) and are being reimplemented in `control/`. What
+follows is therefore evidence rather than documentation: it records what this
+hardware did under a particular rule, so that an implementation can be judged
+against measurements instead of against a description. Where a rule is named it
+is the rule the build under test applied, not one the project ships.
+
+The hardware behaviour is a different matter and does hold generally — a hybrid
+inverter's response to a forcible command, and what a second battery looks like
+to an energy manager, are properties of the equipment.
+
+Sign convention throughout: **positive = charging**, negative = discharging.
+All per-hour figures are hourly means from Home Assistant's long-term
+statistics, not spot readings.
+
+## Installation under test
+
+| | Marstek Venus D | Huawei LUNA2000 |
+|---|---|---|
+| Coupling | AC | **DC (hybrid inverter)** |
+| Usable capacity | 15.36 kWh | 13.8 kWh |
+| Power limit | 2500 W | 7000 W |
+| Discharge floor | 12 % | 5 % |
+| Commandable in daylight | yes | **no** |
+
+Both sit behind one Huawei EMMA energy manager. Two properties of that topology
+drive everything below:
+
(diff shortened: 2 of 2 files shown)


ffunes:
Thanks for the rewrite — the reasoning behind it is right, but the branch has been overtaken since 1 September, so I'm closing this and carrying over the parts that are still live.

**The page is already on the base branch.** ac82c3b landed both `primary-battery-field-evidence.md` and `huawei-emma-registers.md` and linked them in the nav — the documentation half of #335, taken from the source branch. That is what the conflict is: this branch predates it, so merging `mkdocs.yml` as it stands would drop the `Huawei EMMA registers` nav line and leave that page unlinked.

**The preamble is no longer true.** Its central claim is that none of the control-layer behaviour described here is part of Omnibattery and is being reimplemented in `control/`. It has been:

- `control/charge_order.py` — `scarce_solar_day`, measured against the DC-coupled battery's room, reimplemented in 216bbed.
- `drivers/huawei.py` — the daylight guard is in the shipped driver, not proposed.

So the edits that de-attribute those two ("the build under test", "proposed in #381", "Scarcity **is now** measured" → past tense) would now read as less accurate than the text they replace, not more. Your instinct was correct on the day you wrote it; the code caught up in the two days since.

**What I've carried across to `release/v1.5.0`** (1cc7c74):

- The 1 September re-measurement against the #350 fix — two pack handovers with the detector armed, at the same 93 % aggregate where 27 August stalled, neither latched. That is new evidence and the best thing in this PR.
- The line it disproves: "Reported upstream as ffunes/Omnibattery#350 **and not addressed by this work**" → "and fixed there".
- The two section titles. "under the fleet-room threshold" / "under the DC-room threshold" names the threshold each day ran under, which "before/after the fix" never did.

On the recorder: good catch, and worth having for the #381 dawn measurements. Nothing to change in the document — the limitation about 26/27 August having no lux or per-string history stands exactly as written.

PR #397, opened 2026-09-04
feat(pricing): re-plan when the remaining solar forecast moves

syphernl:
## Why

The dynamic-pricing calendar is built once, at 00:05. The only later correction, `_check_dp_pre_slot_reevaluation`, can mark a planned slot skippable — it never adds one. So every input that moves during the day is one-way: the plan can shrink, never grow.

The remaining solar forecast is exactly such an input, and it is the one the whole energy balance rests on. Two failures follow from that.

**A day that turns out cloudier than forecast.** The provider revises the remaining figure down, the deficit the 00:05 balance denied becomes real, and nothing can book the cheap midday hours any more. The battery meets the evening peak near-empty and the house buys it at full price.

**A provider that has not published the new day yet.** Forecast.Solar repopulates its remaining-today figure some time after midnight; at 00:05 it can still read 0 kWh. The balance then concludes the entire day must come from the grid and fills the calendar accordingly. Observed here: a 0.00 kWh forecast at 00:05 on a day that produced normally, with grid charging scheduled across the late morning.

## What

**Re-plan on a forecast move.** `_is_solar_forecast_reeval` follows the excluded-device claim trigger (#341) exactly: bidirectional, 1.5 kWh threshold, 30-minute cooldown, four re-evaluations per day, no trigger after 23:00, and the reference refreshed on every rebuild so it re-arms only after another material move. Wired as phase 2.8 of the dynamic-pricing control handler, next to the claim trigger, and cleared with the rest of the daily state at midnight.

Like the claim trigger, it is deliberately not guarded by `_current_price_slot_active`: the forecast is the input the schedule was built on, so a running slot was planned against solar that no longer exists. Rebuilding may shrink or drop it, and that is the point.

**Defer the daily run on a zero forecast.** With a remaining-today sensor configured, a reported 0 kWh at 00:05 leaves the evaluated date unset and bumps `_dp_eval_retry_count`, so the existing 15-minute retry ladder — already in place for late price data — comes back within the hour. After four attempts it plans anyway: a genuinely broken sensor must not leave the battery empty. Only a reported numeric zero defers; a missing value means the balance never used the remaining sensor, and deferring on that would stall the day for the wrong reason.

## Tests

`tests/test_pricing_engine.py` — nine cases for the predicate and its reference helper: threshold both directions, the evening guard, cooldown, daily cap, and that a refresh disarms the trigger.

`tests/test_dp_solar_forecast_guard.py` — five cases for the daily guard: zero defers, exhausted retries plan anyway, a real forecast plans normally, a missing value plans normally, and no configured sensor plans normally.

Full suite: 2165 passed, 10 skipped.

## Review round

Three defects found reviewing the first commit, fixed in `3dae6a60`.

**A remaining forecast falls all day by itself.** Comparing the reading stored at 00:05 against the live reading made an ordinary sunny morning trip the 1.5 kWh threshold around 08:00, and the daily cap was spent before the afternoon revision the feature exists to catch. The stored reading is now carried forward by the production measured since it was taken (`_daily_solar_energy_kwh`, same-day only), and only the gap against that projection counts as a revision. No same-day accumulator means no projection, and the trigger stays silent.

**The deferral was gated on the `DAILY` horizon** — but the retry ladder re-invokes the evaluation with `REMAINING`, so it deferred once and booked the full day 15 minutes later. The gate is now the first hour of a day that has no plan yet, which covers both horizons; every other `REMAINING` caller runs against a day that already has a plan, and a manual rebuild outside that hour is never held back. `SOLAR_FORECAST_DAILY_RETRY_LIMIT` is 3, so the last ladder call plans regardless.

**An unavailable sensor read as 0.0.** `_remaining_solar_today_kwh` reports zero for an unavailable sensor and for a legacy scalar past the conversion cutoff (`conversion="unsafe_zero"`). That is the right planning input but not a measurement, so a dropout looked like the day collapsing and the reference then stored the phantom zero. Both the trigger and the reference now read through `_read_remaining_solar_reading`, which returns `None` in those cases and keeps the previous reference — the same treatment `_read_excluded_demand_claim_kwh` gives its sensor.

Smaller: the scheduled run clears a retry counter left behind by a day that never completed an evaluation (the control handler's midnight reset is gated on an evaluated date), and the deferral now returns before the diagnostic and curtailment state is torn down.

Full suite after the fixes: 2173 passed, 10 skipped.


ffunes:
Revisado. El diseño calca `_is_excluded_demand_reeval` fielmente y las dos mitades componen bien: si el proveedor publica a la 01:10, después de que la última pasada de la escalera planificó sobre el cero forzado, el disparador de forecast lo recoge (referencia 0 → 18 kWh) y replanifica. Ese es el argumento para llevarlas juntas.

Dos cosas antes de mergear, empujadas a la rama en `c8aca33`:

**La proyección se armaba en instalaciones que no miden producción.** `_read_solar_produced_today_kwh` toma `_daily_solar_energy_date == today` como prueba de que hay datos, pero `handle_daily_energy_reset` pone esa fecha cada medianoche haya fuente o no. Sin `solar_production_sensor` y sin batería con PV (una flota de Venus v3, por ejemplo) el acumulador se queda en 0.0 todo el día: la proyección queda clavada en el forecast de las 00:05 y el declive normal del sensor *remaining* cruza los 1.5 kWh a media mañana. Es el mismo fallo que arregló `3dae6a60`, por otra ruta, y con el cupo diario gastado antes de la revisión vespertina que motiva la feature. `test_solar_forecast_reeval_false_below_threshold` es ese escenario, y pasa solo porque el delta se queda corto. Ahora la proyección exige una fuente real, el mismo test con el que se crea el sensor de solar diaria.

**El día diferido no capturaba la referencia de día completo.** La captura estaba gateada en `horizon is DAILY` y la escalera reinvoca con `REMAINING`, así que en todo día diferido — el caso normal en cuanto el cero difiere — `_daily_solar_forecast_initial_kwh` quedaba sin poner, tanto para el dashboard como para la referencia de aprendizaje del `solar_profile`. Dentro de la primera hora no se ha producido nada, así que la cifra *remaining* es la de día completo; la captura ya guarda solo el primer valor del día.

Un test por arreglo, ambos verificados en rojo sin su cambio. Suite completa: 2320 passed, 10 skipped.

Nota menor que no he tocado: `_read_remaining_solar_reading` solo rechaza `conversion == "unsafe_zero"`, mientras que el hermano en `control/charge_order.py` rechaza además `source in (None, "fallback")`. Hoy son equivalentes porque la rama fallback siempre marca `unsafe_zero`, pero es una trampa si alguna vez divergen.

Gracias por el PR — buen trabajo, sobre todo la ronda de revisión propia.

PR #400, opened 2026-09-04
feat(pricing): keep stored energy for the dearer hours that need it

syphernl:
> **Stacked on #355.** Its two commits are part of this diff until that PR merges into `release/v1.5.0`, after which only `feat(pricing): keep stored energy for the dearer hours that need it` remains here. Review that commit alone.

## Why

`price_discharge` asks whether the current hour is cheap. It never asks whether the dearer hours still ahead need the energy that is in the battery. So a full battery at noon is held out of a mild hour it could have covered, and a nearly empty one at 16:00 is drained into that same hour before the peak it was saved for. Discussions #273 and #270 both ask for the missing half.

## What

An opt-in Dynamic Pricing subfunction. Per rebuild it:

1. takes the price slots between now and local midnight;
2. projects the learned 15-minute consumption profile onto them and subtracts the expected PV, leaving the net grid demand per slot;
3. gives the dearest slots first claim on the energy above the SOC floors, only for slots at least the configured minimum saving above the current price;
4. subtracts the PV expected to refill the battery before that claim falls due, capped by the room left in it;
5. converts the remainder into one percentage of fleet capacity and sets a per-battery `price_reserve` discharge blocker below that level.

Everything above the reserve stays available for self-consumption now, so this composes with the existing thresholds instead of competing with them. The reserve is recomputed every control cycle against the live price and live SOC.

## Why a blocker and not a raised `min_soc`

`_curtailment_battery_snapshots` resolves each battery's floor through `_effective_discharge_min_soc`. Raising that floor would feed this calculation into its own input, and would shrink the pre-discharge budget by the reserve. `price_reserve` is registered as an *economic* blocker, so safety paths, peak shaving, manual time slots and the pre-discharge planner look straight past it.

## Scope and limits

Dynamic pricing only; real-time price mode has no forward curve. Nothing is reserved past midnight — reserving overnight would be wrong on every day the sun refills the battery, and there is no model of tomorrow's PV. The reserve does not know about grid charging planned later in the day, so on a night with a scheduled cheap charge it can hold energy the plan was going to replace.

Refs #273, #270.

## Tests

53 new tests, `2313 passed, 10 skipped`.


ffunes:
Frank — first, thanks. The quality here is genuinely high, the tests are there, and choosing an economic blocker over raising `min_soc` to avoid feeding the reserve back into `_effective_discharge_min_soc` was the right call and not the obvious one. I want to keep getting this.

I do need to change one thing about how we work, though, and it isn't about the code.

I decide the roadmap for this integration, and right now I'm deciding it in response to finished PRs. Four open at once, all in dynamic pricing, means my week goes to evaluating work that already exists instead of building what I think grows the user base. And by the time a PR is written, "no" costs you three days of work — that's not a fair position to put either of us in.

So, from here: **fixes, go ahead, no need to ask.** **Features, get a green light first** — comment on the discussion and wait for me. I've added an `approved` label and I'll start marking discussions with it, so there's a queue you can pull from rather than guessing.

On the PRs themselves: #355, #401 and #397 go into 1.5.0, as they are. I'll merge in that order — #355 first since it moves `config_flow.py` and `pricing/engine.py` the most, then #401, then #397 rebased on top.

**This one waits for 1.6.0.** Two reasons. It reads the DP forward plan, and that plan still can't distinguish the morning slot from the evening one in `_last_decision_data` — I want that fixed before something new depends on it. And it's the second planner reading the same price curve and PV forecast as #355, with no test for the day both are active: `surplus_price_hold` blocking charge while `price_reserve` blocks discharge, battery frozen in both directions, nothing in the panel explaining why. When you rebase off #355, could you add that case?

One correction: `_evaluate_evening_recharge` doesn't call `get_average()`. It's a local of `_evaluate_remaining_grid_charging` (assigned at `engine.py:3867`, called at 3881) and the evening path never references it — same on `main` and `release/v1.5.0`. No `NameError`.

syphernl:
Rebased on `release/v1.5.0` with #355 in. Added `tests/test_surplus_hold_with_discharge_reserve.py`: both planners active on the same day, `surplus_price_hold` blocking charge while `price_reserve` blocks discharge. It asserts both blockers register at once, that each carries its own reason and detail so the panel can attribute the standstill, that releasing either direction leaves the other untouched, and that the reserve only holds a battery that has reached its reserved floor. Full suite green.

PR #401, opened 2026-09-04
feat(predictive): let an excluded device say when its demand is real

syphernl:
## The bug

The #341 remaining-demand claim reads its sensor unconditionally. evcc derives `sensor.evcc_<loadpoint>_charge_remaining_energy` from the vehicle's SOC target and keeps reporting a demand while nothing is plugged in, so an absent car takes solar away from the battery.

Measured on a live instance this afternoon, with no car in the driveway and `binary_sensor.evcc_peblar_business_connected` off:

```
solar_remaining_raw_kwh        3.36
excluded_demand_claim_kwh      3.36   <- the car that is not there
solar_available_to_battery_kwh 0
energy_deficit_kwh             0.87
planned_grid_charge_kwh        0.87
```

The battery then grid-charged 0.87 kWh it did not need. Every evcc user with the claim configured hits this whenever the car is away.

## The fix

An excluded device may name a **presence entity**. While it does not read as present, that device claims nothing.

- A `binary_sensor`, a `device_tracker` or a text status sensor all work: `on`, `true`, `home`, `connected`, `plugged`, `present`, or a state containing a connected/charging keyword in any of the languages the EV detection already covers (`Verbunden`, `Aangesloten`, `Connesso`, `Branché`, `Charging`, …).
- An unavailable, unknown or missing entity counts as absent, so a broken sensor stops reserving rather than reserving for a device that may not be there. The refusal is written to the debug log, so a state nobody anticipated is not silent.
- A gated device contributes a **zero** claim, not *no* claim. `claimable_solar_demand_kwh()` returning `None` means "no usable reading" to its callers, which keep their last reference — that would leave the intraday re-plan blind to a car that has just arrived or left.

Unset on every existing install, where nothing changes.

## Tests

`2175 passed, 10 skipped`. New coverage: the gate itself, the accepted and refused states, a text status sensor per language, a missing entity, one gated device not hiding another device's claim, and both config-flow paths.

Refs #341.


ffunes:
Thanks — the diagnosis is right and the shape of the fix is the one I'd want: an optional gate, unset everywhere today, and `0.0` rather than `None` so the intraday re-plan still sees the car leave. Translation parity is complete (6 languages + `strings.json`, both flow paths). One thing to change before I merge.

### `_PRESENT_SUBSTRINGS` reads every negative state as present

The substring match is symmetric: in every language covered, the negative phrasing *contains* the positive one.

```
'Disconnected'     -> PRESENT  ['connect']
'Nicht verbunden'  -> PRESENT  ['verbund']
'Desconectado'     -> PRESENT  ['conect']
'Niet aangesloten' -> PRESENT  ['aangeslot']
'Non connesso'     -> PRESENT  ['conness']
'Unplugged'        -> PRESENT  ['plug']
'Débranché'        -> PRESENT  ['branch']
'Not present'      -> PRESENT  ['present']
'ontladen'         -> PRESENT  ['laden']    <- via the _CHARGING_SUBSTRINGS union
```

With `binary_sensor.evcc_<loadpoint>_connected` — the entity you measured against — the gate works. But the PR documents a text status sensor as a supported source and the `EntitySelector` offers `sensor`, so anyone pointing at `sensor.evcc_<loadpoint>_charger_status` or a wallbox status entity gets a gate that is permanently open, and #341 keeps grid-charging for the car that is not there. Silently, since the debug log only fires on the refusal path that never runs.

The new tests pass because they only parametrise positive strings.

### Suggested fix: drop the substring pass

Exact match kills every negative at once without a list of negations, and it is less code than what is there now — delete `_PRESENT_SUBSTRINGS` and the `| _CHARGING_SUBSTRINGS` union, and put the localised words in `_PRESENT_STATES` whole:

```python
_PRESENT_STATES = frozenset({
    "on", "true", "home", "connected", "plugged", "plugged in", "present",
    "verbunden", "aangesloten", "connesso", "connecté", "branché",
    "conectado", "connectat", "charging", "cargando", "laden",
})
```

`"disconnected" != "connected"`, and your six existing language cases still pass verbatim. Anything a status sensor reports that this does not know reads as absent, which is the safe direction and is exactly what the docstring already promises.

Could you also add a parametrised test asserting `0.0` for `"Disconnected"`, `"Nicht verbunden"`, `"Desconectado"`, `"Unplugged"`, `"Not present"`? The current negative test only covers machine states (`off`, `not_home`, `idle`), which is why this slipped through.

Minor: `"connect"` appears twice in `_PRESENT_SUBSTRINGS` (the EN entry and the FR/ES/CA one) — moot once the set goes away.

Everything else looks good to me. Green light once that lands.

syphernl:
Fixed in b8cb88ae, exactly as suggested: `_PRESENT_SUBSTRINGS` and the `_CHARGING_SUBSTRINGS` union are gone, the localised words now live in `_PRESENT_STATES` whole. Added a parametrised negative test covering `Disconnected`, `Nicht verbunden`, `Niet aangesloten`, `Desconectado`, `Non connesso`, `Débranché`, `Unplugged`, `Not present` and `ontladen`. Full suite green.

PR #403, opened 2026-09-04
fix: keep the system up when a battery is switched off at start-up

syphernl:
### Problem

A battery that does not answer during setup raises `ConfigEntryNotReady`, which fails the whole config entry: every other battery, the controller and the dashboard go with it, and Home Assistant retries setup in a loop for as long as the device stays off. Someone switching one battery off at its side switch and restarting Home Assistant loses the entire integration until they switch it back on. Nothing detects the unreachable battery either, because the non-responsive tracking only runs once setup has succeeded.

### Change

The unreachable battery is set up unreachable instead of failing the entry. It gets an empty telemetry snapshot and a seeded failure counter, so it appears in `non_responsive_battery_names`, the `non_responsive_batteries` sensor and diagnostics immediately rather than looking healthy.

Adopting it later cannot be done from the coordinator alone: the hardware configuration write, the driver's connect-time entity definitions (model detection, pack discovery) and the first telemetry read all happen in setup, and entities cannot be added to a platform that already finished setting up. So the coordinator schedules one entry reload the first time the battery answers, from either recovery path (a fresh reconnection, or a plain successful poll for a client that reconnects itself). The flag is one-shot; the reloaded runtime re-arms it only if the battery is unreachable again.

Two adjacent failure modes are fixed with it:

- A driver whose `connect()` raises rather than returning `False` (Huawei reads telemetry, ESPHome resolves entities, Zendure/Sessy do aiohttp work) escaped the setup `try` block and left the entry in ERROR with no retry at all. It is now treated as "not there", like a refused connection.
- The interrupted-active-balance migration hands off through the hardware. On an unreachable battery that path fails and latches manual mode into the entry data, freezing the battery once it returns. It now waits for the reload.

Behaviour for a fully reachable system is unchanged.

### Tests

`tests/test_offline_battery_setup.py` covers the deferred reload: armed only for a battery that started unreachable, fired once, carried by a successful reconnection, left armed by a failed one. Full suite: 1957 passed, 10 skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+
+- **One switched-off battery no longer takes the whole system down at start-up**: a battery that did not answer during setup failed the config entry, so every other battery, the controller and the dashboard disappeared with it and Home Assistant retried setup in a loop for as long as the device stayed off — a battery left on its side switch was enough. That battery now starts unreachable while the rest of the system runs, is reported through the non-responsive batteries sensor, and the integration reloads itself once the battery answers again, which restores its hardware configuration write, its model-specific entities and its telemetry. A driver whose connection attempt raises instead of returning a refusal is treated the same way, and the interrupted-active-balance migration waits for a battery it cannot reach rather than latching manual mode on it.
+
 ## [1.4.0] - 2026-09-04
 
 ### Added
--- a/custom_components/omnibattery/__init__.py
+++ b/custom_components/omnibattery/__init__.py
@@ -9394,6 +9394,17 @@ async def _async_migrate_legacy_active_balance(
 
         legacy_notification_config = dict(battery)
         active = _legacy_active_balance_is_running(battery)
+        if active and not getattr(coordinator, "is_available", True):
+            # An interrupted run is handed off through the hardware (idle,
+            # verify, restore), which an unreachable battery cannot do: the
+            # failure path would latch manual mode into the entry and freeze the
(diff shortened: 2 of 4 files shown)


ffunes:
Movido a `release/v1.5.0` (9979a81), que es donde vive el trabajo de la 1.5.0 — el merge a `main` se ha revertido (ef0ecbb) solo por eso, no por el contenido. `main` lo recibirá con el merge de la release.

Dos arreglos encima, en b7ad98c:

- El `disconnect()` de la rama en la que `connect()` lanza puede lanzar a su vez (Hoymiles publica por MQTT en `close()`, Huawei cierra un segundo cliente), y esa excepción escapaba de `async_setup_entry`: entry en ERROR sin reintento, justo el fallo que este PR evita. Ahora va bajo `contextlib.suppress`.
- `reload_entry_when_reachable` es por batería, así que una instalación apagada entera programaba un teardown/rebuild por batería al volver. El marcador vive ahora en el runtime dict de la entry, que setup reconstruye: varias baterías comparten un reload y un corte posterior vuelve a tener el suyo.

Y una nota en el docstring de `_schedule_setup_reload_if_deferred`: todo esto depende de que la batería inalcanzable siga recibiendo entidades en setup, porque son las que suscriben al coordinator y sin suscriptor `DataUpdateCoordinator` nunca programa un poll. Hoy se cumple en todos los drivers (siembran definiciones en `__init__` y connect solo las refina), pero un driver futuro que las resolviera en connect dejaría esa batería colgada hasta un reinicio.

Gracias por el arreglo — el diagnóstico era correcto y el enfoque de reload diferido es el adecuado.

PR #404, opened 2026-09-04
Add weekly full charge grid import with configurable charge power

wouterbouvy:
## Summary

Weekly full charge today only **raises the SOC ceiling to 100%** on the configured weekday. Batteries still need solar surplus or an active predictive grid-charging session to actually draw power — which can leave packs stuck below 100% on cloudy days, during winter or when predictive charging is not running.

This PR adds an **optional, togglable grid-import path** for the weekly 100% cycle, plus a **charge-power slider** to run that import more gently.

### New dashboard settings (Weekly full charge card)

| Entity | Type | Default | Purpose |
|--------|------|---------|---------|
| `switch.weekly_full_charge_grid` | Switch | OFF | Import from the grid on the weekly day until every battery reaches 100% |
| `select.weekly_full_charge_grid_mode` | Select | `immediate` | `immediate` or `solar_first` |
| `number.weekly_full_charge_grid_power_pct` | Number (10–100%) | 100% | Cap charge power during weekly grid import as a % of max charge limits |

**Grid charge modes**

- **`immediate`**: On the weekly day, start regulated grid import as soon as the weekly cycle is active and any automatic battery is below 100%.
- **`solar_first`**: Keep solar-only behaviour first; only start grid import when the solar-forecast energy balance says the sun cannot cover the remaining energy to 100% (reuses `ChargeDelayManager._should_delay_charge(100)`; no/unavailable forecast → allow grid, same fail-safe as charge delay).

**Charge power slider**

- Applies **only** while a weekly-owned grid session is active (`_grid_charge_owner == "weekly"`).
- Scales `_effective_system_capacity()` by the configured percentage so charging is slower and less stressful on the batteries.
- Does not affect predictive grid charging, solar-surplus PD, or the default solar-only weekly path (grid switch OFF).

### Control architecture

- New `WeeklyGridChargeManager` orchestrates session start/stop and is called from `async_update_charge_discharge()` **before** predictive dispatch.
- Grid execution reuses the existing `_handle_predictive_grid_charging()` PD loop (contracted-power regulation, demand protection, rate limiting).
- Weekly grid sessions set `_grid_charge_owner = "weekly"` and per-battery targets of 100%.
- Weekly grid takes priority over predictive charge ceilings on the weekly day while the session is owned.
- Sessions clear on completion, day rollover, or when the grid switch is turned off.

### Observability & docs

- `WeeklyFullChargeSensor` exposes grid mode, grid charging state, owner, and power percentage.
- Weekly status distinguishes `Grid charging` vs `Charging to 100%` (solar-only path).
- Docs updated in `site-docs/features/weekly-full-charge.md`.
- Panel strings added for en/es/ca/de/fr/nl.

## Motivation

Users enabling weekly full charge for cell balancing expect the cycle to **complete reliably** on the chosen day. Without grid import, completion depends on coincidental solar surplus or predictive charging — which is not always available. This keeps the default behaviour unchanged (grid OFF) while giving an explicit, regulated path to finish the weekly 100% cycle from the grid when desired.

## Test plan

- [x] `tests/test_weekly_grid_charge.py` — 11 tests covering:
  - Grid switch OFF → no session
  - `immediate` mode starts weekly-owned grid session with 100% targets
  - `solar_first` defers when forecast says sun is enough; starts when insufficient
  - Weekly grid owner overrides predictive SOC ceiling to 100%
  - Session cleanup on clear
  - Power cap at 100%, 50%, non-weekly owner passthrough, and minimum-floor handling
- [ ] Manual: enable weekly full charge + **Charge from grid** on weekly day; verify regulated import and completion to 100%
- [ ] Manual: try `solar_first` with forecast sensor; confirm grid only starts when forecast gap requires it
- [ ] Manual: set **Grid charge power** to 50%; confirm slower charge rate during weekly grid session only


wouterbouvy:
This fixes: https://github.com/ffunes/Omnibattery/discussions/357

ffunes:
Thanks for this, Wouter — the problem you identified is real and your write-up is what made me go and look at why it happens. I'm not going to merge this one, though, and I want to explain properly rather than just closing it.

My objection isn't the code quality, it's the shape of the solution: a new switch, a new select and a new number slider, plus a second grid-charge owner running alongside the predictive one. Every setting we add is a decision the user has to understand and get right, and this one asks them to reason about grid modes and charge-power percentages just to make a feature they already enabled actually finish. The integration already has too many knobs; I'm trying hard to take them away, not add more. Two owners competing for the grid-charge path is also a class of bug I'd rather not sign up for.

The root cause turned out to be narrower than it looks. Predictive charging *already* knows how to buy energy from the grid, in every mode including flat tariffs — they all route through the same `_should_activate_grid_charging()`. It never helped the weekly cycle for two reasons:

1. That balance answers "will I run out of battery", not "is the battery full". Only forecast consumption and the guaranteed minimum SOC could produce a deficit. A weekly 100% day is neither, so the answer was always "no charge needed".
2. Every ceiling in the planning chain was `max_soc` — the energy balance headroom, the per-battery grid targets, the dispatch limit. The top slice of the pack didn't even exist as chargeable space, so raising the ceiling alone changed nothing.

So instead of a second charging path, the weekly gap now enters that same balance, exactly the way the guaranteed minimum SOC does, and the ceiling rises to 100% while a weekly cycle is pending. The gap is netted against the solar surplus the forecast expects, so the grid only buys what the sun won't deliver — which also gives you your `solar_first` behaviour for free, without a mode to choose. It works on flat tariffs too, because it's the same evaluation for all predictive modes.

Net result: your use case, zero new settings, ~90 lines. It's on `release/v1.5.0` (1a23b20) if you want to look at it, and I'd genuinely value your review of it — you clearly understand this part of the codebase.

The one thing I did not carry over is the charge-power slider. It's a reasonable idea on its own merits (gentler top-of-charge), but it belongs to the weekly/taper feature rather than to a grid path, and I'd want to see evidence it matters before adding another number entity. If you have measurements suggesting it does, open an issue and I'll look at it seriously.

Thanks again for the time you put into this, and please keep them coming.

wouterbouvy:
Thanks for your review and elaborate answer @ffunes!

> My objection isn't the code quality, it's the shape of the solution: a new switch, a new select and a new number slider, plus a second grid-charge owner running alongside the predictive one. Every setting we add is a decision the user has to understand and get right, and this one asks them to reason about grid modes and charge-power percentages just to make a feature they already enabled actually finish. The integration already has too many knobs; I'm trying hard to take them away, not add more. Two owners competing for the grid-charge path is also a class of bug I'd rather not sign up for.

I fully agree on this. Less is more and altough adding extra buttons/sliders gives more configurability, it also makes the learning curve steeper.

> The root cause turned out to be narrower than it looks. Predictive charging _already_ knows how to buy energy from the grid, in every mode including flat tariffs — they all route through the same `_should_activate_grid_charging()`. It never helped the weekly cycle for two reasons:
> 
> 1. That balance answers "will I run out of battery", not "is the battery full". Only forecast consumption and the guaranteed minimum SOC could produce a deficit. A weekly 100% day is neither, so the answer was always "no charge needed".
> 2. Every ceiling in the planning chain was `max_soc` — the energy balance headroom, the per-battery grid targets, the dispatch limit. The top slice of the pack didn't even exist as chargeable space, so raising the ceiling alone changed nothing.
> 
> So instead of a second charging path, the weekly gap now enters that same balance, exactly the way the guaranteed minimum SOC does, and the ceiling rises to 100% while a weekly cycle is pending. The gap is netted against the solar surplus the forecast expects, so the grid only buys what the sun won't deliver — which also gives you your `solar_first` behaviour for free, without a mode to choose. It works on flat tariffs too, because it's the same evaluation for all predictive modes.

This is a great solution I think. Only caveat is that forecast predictions can be off and too less energy is bought. The final % to top of the batteries is still depending on solar surplus. This can be mitigated by adjusting the predictive charging parameters maybe. But in order to calibrate SoC correcly it's off importance to reach that 100%.

> Net result: your use case, zero new settings, ~90 lines. It's on `release/v1.5.0` (1a23b20) if you want to look at it, and I'd genuinely value your review of it — you clearly understand this part of the codebase.
> 
> The one thing I did not carry over is the charge-power slider. It's a reasonable idea on its own merits (gentler top-of-charge), but it belongs to the weekly/taper feature rather than to a grid path, and I'd want to see evidence it matters before adding another number entity. If you have measurements suggesting it does, open an issue and I'll look at it seriously.

I have no measurements myself. But research show that a lower C-rate results in lower current draw or charge rate relative to the battery's total capacity, which significantly reduces physical and thermal stress on the cell.
I don't mind stressing the cells more if I want to capture PV, but when I need to import from the grid I would like to have the choice to charge on a slower rate.
Maybe this is something to consider to add to the predictive charging dashboard, as it now fits better there.
 
> Thanks again for the time you put into this, and please keep them coming.

You are the one that should thanked for your great work!

PR #405, opened 2026-09-04
Add Anker battery health telemetry and dashboard refinements

wouterbouvy:
## Summary

- Exposes **Anker battery state of health (SoH)** from Modbus input register **10015** as `sensor.<battery>_battery_soh` for all supported Solarbank models, with `(SoH)` in entity names across all locales.
- Shows SoH on the Omnibattery dashboard battery card under **Health & cells** when the sensor is available.
- Makes dashboard health metrics **entity-driven**: voltage, cell min/max and cell-delta rows are rendered only when matching sensors exist, so Anker cards no longer show empty placeholders for Modbus readings the driver does not provide.
- Documents the feature in Anker battery guides (en/es), entity reference (en/es) and `CHANGELOG.md`.

Field verification note: SoH has been confirmed on **Solarbank Max AC** (product code **DMWH**). Other Solarbank models share the register map but are not yet confirmed in the field.

## Motivation

Anker users currently have no native SoH visibility in Omnibattery, while the dashboard always reserved rows for voltage and per-cell metrics that Anker does not expose over Modbus. This adds the useful health reading and trims the card to what each brand actually provides.

### Driver & entities
- `drivers/anker.py`: register **10015** field spec + `battery_soh` sensor (`mdi:battery-heart`, precision 1, low scan interval).
- Translations/strings for `battery_soh` in en/de/es/fr/nl/ca.

### Dashboard
- `marstek-panel.js`: conditional SoH metric; voltage/cell metrics gated on entity presence; localized **State of health (SoH)** label.

### Docs
- `site-docs/configuration/batteries/anker.md` + `.es.md`: diagnostics, dashboard behaviour, DMWH verification note.
- `site-docs/reference/entities.md` + `.es.md`: `battery_soh` row; Anker omissions noted for voltage/cell sensors.
- `CHANGELOG.md`: Unreleased entries.

### Tests
- `tests/test_anker_driver.py`: decode from register 10015; sensor definition coverage for DN7M and DMWH.

## Test plan

- [x] `tests/test_anker_driver.py` — 48 passed locally (includes new SoH decode/definition tests)
- [ ] Manual: Anker Solarbank Max AC — verify `sensor.<battery>_battery_soh` reports plausible % and appears on dashboard **Health & cells**
- [ ] Manual: confirm Anker card no longer shows empty Spanning / cell min / cell max / Δ cel rows
- [ ] Manual: Marstek/Zendure battery card still shows voltage and cell metrics when entities exist


wouterbouvy:
image

Looks like this now

ffunes:
Thanks for this @wouterbouvy — the feature is good and I want it in 1.5.0. I've rebased it onto `release/v1.5.0` and opened #413 to avoid holding it up; your four commits are cherry-picked unchanged and authorship is preserved on all of them.

### What was wrong

Not the code — the base. This branch was cut from `main` after 1.4.0 shipped, but `release/v1.5.0` forked from `main` earlier, at `81971f0` (2026-09-01, manifest `1.4.0b8`). So the merge base predates both, and the three-dot diff pulled in 10 unrelated commits of mine — 39 files, +1577/-345, instead of your actual 15 files / +154.

`git merge-tree` gave 13 conflicts, including `add/add` on `control/pack_soc.py` and `tests/test_venus_ad_pack_soc.py`. Resolving those by hand would have risked reverting #408, since `release/v1.5.0` is ahead of `main` in that file. Rebasing made 12 of the 13 disappear — they were artefacts of the stale merge base, not real. Only `CHANGELOG.md` genuinely conflicted, because `release/v1.5.0` already has its own `[Unreleased]` block; your entry now sits under the existing `### Added` plus a new `### Changed`.

For next time: `git rebase --onto upstream/release/v1.5.0 83c6a48b <your-branch>`.

### One thing I'd like your call on

`battery_soh` is created for every Solarbank, but you've only field-verified `DMWH`. In `read_telemetry` the field is skipped only when the whole block read fails or comes back short — and 10015 sits inside `(10000, 10050)`, which is already read in full. So on a model that doesn't implement the register but answers with `0` rather than an exception, the sensor would publish `0.0 %`, which reads as a dead battery rather than "no data".

Two ways out, and you're better placed to pick: gate the sensor on confirmed product codes (same shape as `_INDEPENDENT_PV_PRODUCT_CODES` right next to it), or map `0 -> None`. I don't have Anker hardware and I don't know this driver the way you do, so I'd rather not guess at what those units actually return. If you know an unsupported model throws instead of answering zero, then there's nothing to fix and I'll merge as is.

### Minor, non-blocking

- `precision: 1` on a `uint16` renders `92.0 %`; `0` reads better. Same in the panel (`this._nf(b.soh, 1)`).
- `scan_interval: "low"` is inert here — `_build_read_groups` derives cadence from batch ranges, not per-sensor definitions. No harm, just doesn't do what it looks like.
- `_batSig` only signs the device set, so the new entity gating is evaluated once when the card is built. If that happens before the registry is populated, the rows stay hidden until the panel reloads.
- The `### Changed` entry says "Anker", but the gating affects every brand — a Marstek with the balance monitor off loses its Δ cell row too.
- `"Battery State of Health (SoH)"` with `has_entity_name = True` renders as "<device> Battery State of Health (SoH)".

None of these are in #413; I deliberately left your commits untouched. Happy to take them as a follow-up from you, or fold them in myself if you'd rather.

Closing this one in favour of #413 — please review there.

PR #407, opened 2026-09-04
fix(setup): one reload for an offline fleet, and never let close() fail the entry

ffunes:
Follow-up to #403, from reviewing it after the merge.

### `close()` on a link that never came up

`connect()` raising is now treated as "the battery is not there", and that path calls `coordinator.disconnect()`. But a close can raise in its own right — the Hoymiles driver publishes over MQTT when `_connected` is set, Huawei closes a second client — and that exception escaped `async_setup_entry` uncaught, leaving the entry in ERROR with no retry. That is exactly the failure #403 exists to prevent, reachable through its own recovery path. Suppressed.

### One reload for a fleet that was switched off together

`reload_entry_when_reachable` is per battery, so three batteries coming back in the same poll cycle scheduled three `async_schedule_reload` calls: three sequential teardown/rebuilds of the same entry, where one adopts all of them.

The marker now lives in the entry's runtime dict. Setup rebuilds that dict, so the reloaded runtime starts unmarked and a later outage still gets its own reload.

### Note on the recovery path

Documented on `_schedule_setup_reload_if_deferred`: the whole mechanism depends on the unreachable battery still getting entities at setup, because entities are what subscribe to the coordinator and `DataUpdateCoordinator` never schedules a poll without a subscriber. Every driver seeds its definitions in `__init__` today (connect only refines them), so this holds — but a future driver that resolved its entities at connect would leave a battery that started unreachable stuck until a restart.

### Tests

Two added to `tests/test_offline_battery_setup.py`: batteries returning together share one reload, and the marker does not survive into the reloaded runtime. Full suite: 1959 passed, 10 skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


ffunes:
Moved onto `release/v1.5.0` instead (b7ad98c), where the rest of the 1.5.0 work lives. `main` picks it up with the release merge.

PR #408, opened 2026-09-04
A Venus D empties at its first pack, not its fullest

sphings79:
Follow-up to the per-pack work you asked me to test on #350 — I have it running now, and it found something the same evening.

## What the hardware does

`soc_vs_floor` returns `max(pack_soc)`: empty only once the *fullest* pack reaches the cutoff. A Venus D does the opposite. It stops the whole battery the moment its **first** pack reaches the cutoff, and leaves the charge in the others where it is. It could keep going on the remaining packs; it does not.

Measured here on 4 September, cutoff at 12 %:

| | | | | | |
|---|---|---|---|---|---|
| pack 1 | 12.1 % | pack 4 | 12.0 % | aggregate | **15 %** |
| pack 2 | 12.0 % | pack 5 | 20.3 % | device usable SOC | **3.4 %** |
| pack 3 | 19.0 % | pack 6 | 19.3 % | delivered | **0 W** |

`max(pack_soc)` reads 20.3 % there and says keep going. So the control layer commanded **−938 W into a battery that had already stopped, for six hours**, never listed it among `battery_discharge_blockers`, and the house imported 1229 W from the grid while the other battery ran itself down to its own floor. PD quality sat at `sluggish` with 1206 W of RMS error, which was exactly that.

The aggregate hid it as well: 15 % against a 12 % floor reads perfectly healthy.

Pack SOCs read from a separate Modbus integration on the same registers, so they are independent of the driver's own probe.

## The change

`min(pack_soc)` on the floor. The ceiling stays `min(pack_soc)` and is untouched.

The two ends are not mirrors, because the hardware is not symmetric: charging *does* walk on to the next pack when one fills — that is what the four #350 handovers showed — while discharging does not walk on at all.

This also restores what `tests/test_venus_ad_pack_soc.py` already says in its own module docstring:

> the *additional* discharge floor: a pack at min_soc blocks discharge even when the aggregate is still above it. The symmetric swap (max(pack_soc) replacing the aggregate, which would extend discharge) is deliberately not implemented

That is the behaviour this PR implements; the code had the swap the docstring rules out. The three floor tests pinned the swapped version, so they are rewritten, and the measured six-pack case is added as its own test.

## What it does not do

It does not recover the stranded charge — roughly a kilowatt-hour sitting in packs 3, 5 and 6. That is Marstek's choice and nothing here reaches it. What changes is that the control layer stops allocating a discharge this battery will not serve, so the share goes to one that can. On the evening above it would have changed nothing, because the hybrid was at its own floor too; on an evening with headroom it would.

Suite: 2303 passed, 10 skipped.


ffunes:
Reviewed: pulled the branch, ran the suite, and tested the one case I was worried about.

## The direction is right

The original commit (e521d12) contradicts itself. Its own test module docstring says:

> the *additional* discharge floor: a pack at min_soc blocks discharge even when the aggregate is still above it. The symmetric swap (max(pack_soc) replacing the aggregate, which would extend discharge) is deliberately not implemented — see the plan's F3

and the code shipped exactly the swap the plan ruled out, with tests pinning it. Your six-pack measurement closes it: floor 12 %, aggregate 15 %, delivering 0 W — neither `max(pack)` (20.3 %) nor the aggregate (15 %) is the number the device acted on; `min(pack)` (12.0 %) is. The change goes in.

## Blocker: an absent slot leaves a stale `0` in `coordinator.data`

`_learn_pack` (`drivers/marstek.py:506`) documents that a slot the hardware does not have "reads a flat 0". During the three probe cycles **every** slot is polled, so that `0` goes into `updated_data` and lands in `self.data.update(...)` (`infra/coordinator.py:1413`). When the probe finishes the driver stops polling the slot and hides the entity — but it **never removes the `0` from `coordinator.data`**. `pack_socs()` reads by key prefix and `0 <= 0 <= 100` clears the range guard.

Under `max()` that was harmless. Under `min()` it is fatal:

```python
# 2-pack Venus D, both real packs at 80 %.
# Slots 3..6 answered a flat 0 during the probe, were written off, and the 0 stuck.
data = {
    "battery_soc": 80,
    "battery_soc_pack_1": 80.0, "battery_soc_pack_2": 80.0,
    "battery_soc_pack_3": 0.0, "battery_soc_pack_4": 0.0,
    "battery_soc_pack_5": 0.0, "battery_soc_pack_6": 0.0,
}
```

```
release/v1.5.0 (max):  floor_soc = 80.0  -> dischargeable        PASS
this branch    (min):  floor_soc =  0.0  -> excluded from discharge   FAIL
```

And it does not recover: `__init__.py:4092` sets `_discharge_min_soc_latched = True`, which only releases on `floor_soc >= min_soc + DISCHARGE_MIN_SOC_REENTRY_MARGIN`, and `floor_soc` never leaves 0. A full battery drops out of the discharge fleet permanently.

This hits any Venus A/D with fewer than six populated slots on firmware that answers 0 for empty ones. Your unit has all six populated, which is why the measurement did not show it.

**Suggested fix, in the driver rather than here:** on the cycle the probe finishes, emit `snapshot[key] = None` once for the written-off slots. The coordinator already stores `None` as-is (`coordinator.py:1392`) and `pack_socs()` filters it out on the `isinstance` check. Two lines in `_learn_pack`, `control/pack_soc.py` untouched. Filtering `> 0` inside `pack_socs()` would also work but it masks stale data instead of purging it.

## Minor

- **No CHANGELOG entry.** e521d12 added one; this should too.
- **A dropped test with no replacement.** `test_packs_override_an_aggregate_already_at_the_floor` pinned the other direction — aggregate at the floor while every pack is above it, so keep going. That behaviour survives `min()` unchanged (verified: `soc_vs_floor` returns 12.0 with the aggregate at 10) and is now uncovered. Keep it with the same data and the same expectation.
- `test_one_pack_at_the_floor_ends_the_discharge` and `test_an_aggregate_above_the_floor_does_not_override_a_pack_on_it` are the same case twice. Cosmetic.
- One note on the evidence: `device usable SOC 3.4 %` is exactly `(15 − 12) / (100 − 12)`, i.e. derived from the aggregate, not from the packs — so on its own it does not demonstrate the "stops at its first pack" mechanism. The 0 W delivered with the aggregate three points clear of the floor does rule out both `max()` and the aggregate, so the conclusion holds; the reasoning just leans on a number that is not saying what it appears to say.

`tests/test_venus_ad_pack_soc.py`: 20 passed on the branch.

Fix the stale-zero case and this is good to merge.

sphings79:
Blocker fixed, minors done, and one correction to my own evidence — you were right about it.

## The stale zero

Fixed your way, in the driver. `_learn_pack` now overwrites the key with `None` on the cycle a slot is written off, so the coordinator's stored zero is replaced rather than left behind. `control/pack_soc.py` is untouched, and the `isinstance` filter drops it. Filtering `> 0` there would have hidden the stale value instead of removing it, and left the same zero showing in the entity and in diagnostics.

Two tests: the state you described — two real packs at 80 %, four slots holding `None` — stays dischargeable, and a probe-level test asserting the purge happens exactly once, on the cycle the probe finishes, names every written-off slot and never touches a populated one.

**Your reading of the hardware is confirmed here, incidentally.** This unit's map exposes a seventh slot at register 34602, and it reads a flat 0 beside six populated packs reporting real values. So "an absent slot answers 0" is not a documented guess — it is what this firmware does, on the same battery the fix is for.

## The correction

You are right that `device usable SOC 3.4 %` proves nothing about packs: it is `(15 − 12) / (100 − 12)`, derived from the aggregate. I had noticed the same relationship in the raw figure (`0.46 kWh` = 3 % of 15.36 kWh) and then quoted it anyway as though it were independent evidence. It is not, and I should not have leaned on it.

What carries the conclusion is the part you name: 0 W delivered with the aggregate three points clear of the floor rules out both the aggregate and `max(pack)`. The mechanism itself comes from the owner, who confirms the unit stops at its first pack and strands the rest — not from that number.

The docstring in `pack_soc.py` quotes the measurement, so I have left the figures in but they now sit as the observation they are.

## Minors

- **CHANGELOG** entry added under Fixed.
- **`test_packs_override_an_aggregate_already_at_the_floor` restored**, same data, same expectation. You are right that it survives the swap: `soc_vs_floor` returns 12.0 with the aggregate at 10.
- **The duplicate dropped.** `test_an_aggregate_above_the_floor_does_not_override_a_pack_on_it` was the first-pack case again; the remaining one keeps the clearer name.

## A lead, not part of this PR

Register **30212** on this unit is labelled `pack_config` in a community map and reads **5** with six packs installed. Zero-based that is exactly six, which would make the whole probe unnecessary — no cycles, no write-off, no stale zeros to purge.

One installation and one value, so I am not proposing anything on it. The owner has offered to physically remove a pack and see whether it reads 4. If it does I will open that separately; this PR should not wait on it.

Suite: 2305 passed, 10 skipped.

PR #413, opened 2026-09-06
Add Anker battery health telemetry and dashboard refinements (rebase of #405)

ffunes:
Rebase of #405 onto `release/v1.5.0`. All code is @wouterbouvy's, authorship preserved on every commit.

## Why a new PR

#405 targets `release/v1.5.0` but was branched from `main` after 1.4.0 was released. `release/v1.5.0` forked from `main` at `81971f0` (2026-09-01, manifest `1.4.0b8`), before 1.4.0 was closed, so the merge base is older than both branches and the three-dot diff dragged in 10 unrelated commits — 39 files, +1577/-345.

`git merge-tree` gave 13 conflicts, including `add/add` on `control/pack_soc.py` and `tests/test_venus_ad_pack_soc.py`. Resolving those by hand risked reverting #408 (`release/v1.5.0` is ahead of `main` there by ~104 lines).

Rebased with `git rebase --onto release/v1.5.0 83c6a48b`. Three commits applied clean; only `CHANGELOG.md` conflicted, because `release/v1.5.0` already has its own `[Unreleased]` block — the Anker entry now sits under the existing `### Added` and a new `### Changed`.

Result: **15 files, +154/-33**. `2340 passed, 10 skipped`.

## What it does

- Exposes Anker **state of health** from Modbus input register **10015** as `sensor.<battery>_battery_soh` for all supported Solarbank models. 10015 already falls inside the polled `(10000, 10050)` input batch range, so there is no extra bus traffic.
- Makes the dashboard **Health & cells** metrics entity-driven: voltage, cell min/max and cell-delta rows render only when matching sensors exist, so Anker cards stop showing empty placeholders.
- Docs (en/es) and translations for all six locales plus `strings.json`.

Field-verified on Solarbank Max AC (`DMWH`). Other product codes share the register map but are unconfirmed.

## Review notes (not addressed here — deliberately kept to @wouterbouvy's original commits)

- `precision: 1` on a `uint16` register renders `92.0 %`; `0` would be right. Same in the panel (`this._nf(b.soh, 1)`).
- `scan_interval: "low"` on the sensor definition is inert for Anker — `_build_read_groups` derives cadence from batch ranges, not per-sensor.
- `_batSig` only signs the device set, so the new entity gating is evaluated once at card-build time. If the card is built before the registry is populated, the rows stay hidden until the panel reloads.
- The `### Changed` entry says "Anker", but the gating affects every brand: a Marstek with the balance monitor off (no `cell_delta`) also loses that row.
- `"Battery State of Health (SoH)"` with `has_entity_name = True` renders as *"<device> Battery State of Health (SoH)"*.
- Docs file it under "Diagnostics" but the sensor carries no `category: "diagnostic"`.

Closes #405.


wouterbouvy:
**Your call — SoH on unverified models:** I kept the sensor on all Solarbank models sharing the register map and map register value **0 → unavailable** in `read_telemetry`, so units that don't implement 10015 won't show a false 0 %. Trade-off: a genuine 0 % SoH would also display as unavailable; I'm OK with that given how rare that is on verified hardware.

**Also addressed:**
- SoH `precision: 0` in driver + panel
- Removed inert `scan_interval` from Anker SoH definition
- Extended `_batSig` to rebuild cards when layout-affecting entities appear
- Changelog/docs: entity-driven gating is cross-brand; docs section renamed to Health telemetry
- Entity naming left as-is for consistency with Battery SOC

Follow-up PR **#417** targets `feature/anker-soh-v150` with the two cherry-picked commits (`49a1011`, `13d6f72`). Merge that into this branch when ready.

wouterbouvy:
Opened follow-up **#417** (`wouterbouvy:feature/anker-soh-v150-feedback` → `feature/anker-soh-v150`) with the review-feedback commits cherry-picked on top of this PR's branch. Merging #417 will update #413 automatically.

Summary of what's in #417:
- `0` from register 10015 → unavailable (SoH on all Solarbank models; no product-code gate)
- SoH precision `0`, removed inert `scan_interval`, `_batteryCardSig()` for panel rebuild
- Health telemetry docs + cross-brand changelog wording

PR #440, opened 2026-09-10
Add Zonneplan hourly and quarter-hourly dynamic pricing support

hprax:
## What

Add **Zonneplan** as a Dynamic Pricing source using the existing [Zonneplan One integration](https://github.com/fsaris/home-assistant-zonneplan-one). Users can select an hourly, quarter-hourly or legacy electricity tariff sensor directly, without a template sensor or additional API credentials.

This is a single Zonneplan commit based on `release/v1.5.0` (`a11dc8ab`), with the unrelated upstream-main commits removed. The implementation uses the release branch's shared configuration and pricing helpers.

## Approved roadmap item

No separately approved roadmap issue is linked. This revision addresses the maintainer's [review of this PR](https://github.com/ffunes/Omnibattery/pull/440#issuecomment-5655015704).

## Implementation

- Register `PRICE_INTEGRATION_ZONNEPLAN` in `calculations.PRICE_LIST_ATTRS` and `parse_prices_for_integration()`. Both import and optional export curves use this dispatch; `engine.py` requires no changes.
- Add Zonneplan once to `_price_integration_options()` and once to `_validate_price_sensor()`. Initial setup, options and export selection inherit the shared behavior.
- Parse modern `start_date` / `end_date` / `price_tax_included.amount` forecasts, and legacy `start_date` or `datetime` / `electricity_price` forecasts.
- Convert forecast amounts to tax-inclusive EUR/kWh by dividing by **10,000,000**. Current sensor states already use EUR/kWh and are not scaled again. Zero and negative prices are preserved.
- Preserve explicit modern interval boundaries. Legacy entries last one elapsed hour; missing entries do not stretch an earlier price across a gap.
- Reject incompatible units, invalid forecast containers, non-finite or boolean prices and non-positive intervals. Keep valid entries when another entry is malformed; sort and deduplicate matching local intervals. Per-entry exception logging includes the offending entry and error.
- Add labels in all existing translations, matching English and Spanish configuration documentation, and an entry under `1.5.0b2 → Added` with contributor attribution.

The shared planner's energy-deficit, efficiency, SOC and power-limit calculations are unchanged. Zonneplan is also selectable for the existing optional export curve, but this PR does not derive an export tariff from an import tariff: the configured sensor must represent the user's actual applicable price.

## Modern hourly format: live verification

On 2026-09-13, the modern **Current hourly electricity tariff** sensor was temporarily enabled in an existing Home Assistant installation. This is a real `forecast` entry from that sensor, distinct from the legacy hourly and quarter-hourly sensors:

```yaml
- start_date: '2026-09-12T21:00:00+02:00'
  end_date: '2026-09-12T22:00:00+02:00'
  price_tax_included:
    amount: 3733562
  price_tax_excluded:
    amount: 2625081
  tariff_group: normal
  sustainability_score:
    permille: 566
```

It includes `end_date` and parses to one hourly slot at **0.3733562 EUR/kWh**. This exact entry is now a regression test. The sensor was restored to disabled after inspection. Legacy hourly and modern quarter-hourly attributes were also previously verified in the same installation.

Modern entries without an end remain invalid rather than guessing whether they represent 15 or 60 minutes; this is covered by the malformed-entry tests.

## Tests

Local environment: Windows, Python 3.12, repository test dependencies.

```text
python -m pytest -q --disable-warnings
2572 passed, 10 skipped, 8 warnings

python -m pytest -o addopts="" tests/test_config_backup.py tests/test_domain_migration.py tests/test_system_uid_migration.py -q --disable-warnings
11 passed
```

The default run skips tests requiring the Home Assistant plugin; the second command runs the workflow's dedicated flow-test files with that plugin enabled. The 31 Zonneplan cases cover scaling, hourly/quarter-hourly and legacy formats, missing/invalid data, gas rejection, sorting, gaps, negative-price selection, timezone conversion, clock rollback, current state, shared dispatch and stringified-attribute diagnostics, independent export curves, shared export validation, setup/options persistence and the real modern hourly sample.

Python syntax and integration JSON validation pass, and `git diff --check` is clean. No installation or physical battery-control test was performed for this revision.

## Known limitation

The existing planner uses naive local datetimes. Repeated local intervals during autumn rollback cannot be represented separately: identical boundaries are deduplicated and intervals with a non-positive local duration are omitted. Spring-forward durations likewise inherit the shared wall-clock limitation. This provider follows that existing convention and does not claim complete DST scheduling correctness.

## Configuration

Select **Zonneplan** and the electricity tariff sensor matching the contract. Enter thresholds in EUR/kWh, without adding taxes a second time. Tariff-group and cheapest-hour sensors are not forecast sources. Existing provider configurations do not require migration.


ffunes:
Thanks for this — the arithmetic is right and the tests are well built. I checked the scale against `zonneplan_one/const.py` myself: `value_factor: 0.0000001` for both `price_per_date_and_hour.*.electricity_price` and `price_per_date_and_quarter_hour.*.price_tax_included.amount`, unit `€/kWh`, attribute `forecast`. All of that matches what you implemented.

The problem is the base. The PR comes from your fork's `main`, which carries 18 upstream commits `release/v1.5.0` doesn't have — so GitHub reads it as 1866/325 for what is really a 197-line commit, and it's conflicting. More importantly, the branch it was written against has since been refactored, and the integration points you patched no longer exist:

**1. `engine.py` — both hunks land in dead code.** On `release/v1.5.0` the inline `elif` chain in `_parse_price_data` and `_PRICE_LIST_ATTRS` are gone. Import *and* export both route through `calculations.parse_prices_for_integration` (engine.py:945) and `calculations.PRICE_LIST_ATTRS` (engine.py:933). As written, after a conflict resolution Zonneplan falls through to the Nordpool fallback and parses nothing. Please move the two additions there instead:

```python
PRICE_LIST_ATTRS = {
    ...
    PRICE_INTEGRATION_ZONNEPLAN: ("forecast",),
}

def parse_prices_for_integration(integration_type, attrs):
    ...
    if integration_type == PRICE_INTEGRATION_ZONNEPLAN:
        return parse_zonneplan_prices(attrs)
```

Two lines, and the export/feed-in curve works for free — which it wouldn't with the current placement.

**2. `config_flow.py` — four duplicated hunks against the old shape.** The base now has `_price_integration_options()` and `_validate_price_sensor()` shared by setup, options, import and export. Adding `PRICE_INTEGRATION_ZONNEPLAN` to the options list plus one branch in the validator replaces all four blocks.

**3. Missing CHANGELOG entry**, under `## [1.5.0b2] → ### Added`, one line ending in "Thanks to @hprax".

**4. Missing `site-docs/configuration/predictive-charging/dynamic-pricing.es.md`** — every site-doc page has a Spanish sibling.

**One thing to confirm before merge:** the *modern hourly* format. If `price_per_hour` entries carry `price_tax_included` but no `end_date`, the parser `continue`s on every entry and the provider yields zero slots silently. You verified quarter-hourly and legacy against a live install, but modern hourly only against the source. Could you paste a real `forecast` entry from the hourly sensor?

Nit: the per-entry debug log doesn't include the offending entry, unlike `parse_ckw_prices` / `parse_epex_prices`.

Could you rebase just the `Add Zonneplan dynamic pricing source` commit onto `release/v1.5.0` and rework the above? The diff should come out around 60 lines of code plus tests, changelog and the ES doc.

hprax:
Reworked on top of `release/v1.5.0` (`a11dc8ab`) as a single Zonneplan commit; the 18 unrelated upstream-main commits are no longer part of the diff.

- Moved provider registration and dispatch into `calculations.PRICE_LIST_ATTRS` and `parse_prices_for_integration()`. There is now no `engine.py` diff; added regressions exercise both import and independent export curves.
- Replaced the duplicated setup/options changes with one shared options-list entry and one `_validate_price_sensor()` branch, also used for export.
- Added the `1.5.0b2 → Added` changelog line with attribution and the Spanish documentation sibling.
- Included the offending entry and exception in the per-entry debug log.

I also checked the **modern hourly** sensor (sensor.zonneplan_current_hourly_electricity_tariff) on the live Home Assistant installation, a real `forecast` entry is:

```yaml
- start_date: '2026-09-12T21:00:00+02:00'
  end_date: '2026-09-12T22:00:00+02:00'
  price_tax_included:
    amount: 3733562
  price_tax_excluded:
    amount: 2625081
  tariff_group: normal
  sustainability_score:
    permille: 566
```

`end_date` is present; the parser returns a one-hour slot at `0.3733562 EUR/kWh`. That exact sample is now tested. Modern entries missing their end still fail validation rather than guessing the interval length.

Validation: **2,572 passed, 10 skipped** in the full default suite; **11 passed** in the separate Home Assistant flow-test run used by the workflow. Python syntax, JSON validation and `git diff --check` also pass.

The revised commit is available on `hprax:zonneplan-release-v1.5`. This existing PR still uses `hprax:main` to preserve its number and review discussion, so that branch points to the same revised commit.

iceicerickert:
this is great, i built one myself and use the CKW format for now. 
But this is better?
Any "forecast" for releasing this one? :) 
Maybe also an idea to have a custom one? Like CKW format maybe but where you can choose your own currency.

ffunes:
Merged — thanks, this is exactly what I asked for and then some.

The rebase came out at 259/-2 across 14 files with zero `engine.py` diff, which is the right shape: the two lines in `calculations.py` give the export/feed-in curve for free, and the shared `_price_integration_options()` + one `_validate_price_sensor()` branch replaced all four duplicated config_flow blocks. The live modern-hourly `forecast` entry settled the one thing I couldn't verify from the source — good call turning it into a regression test rather than just pasting it here.

I moved the CHANGELOG line to `[Unreleased]` on merge; `1.5.0b2` had already been cut by the time this landed. Nothing for you to do.

Thanks for the patience with the rebase request.

ffunes:
@iceicerickert thanks — Zonneplan is merged now, so it'll be in the next beta.

Whether it's "better" than your CKW setup depends on your contract: if you're a Zonneplan customer it reads your actual tariff directly, no template sensor. If you're not, CKW stays the right choice.

On the generic provider with a configurable currency: open a feature request with the specifics — which attribute shape you'd want to point it at, how the currency should be picked, and what the threshold fields should show. There's a real question there about how many formats one "custom" parser can absorb before it's worse than a per-provider one, so the requirements matter. I'll look at it once it's written up.

PR #448, opened 2026-09-11
Solar charge delay: Calculating the charging duration based on the forecasted yield

gojux:
## What

When calculating the required charging duration, the maximum rated charging power was used, which resulted in unrealistically short charging estimates and caused the charging process to start too late.

With this change, the forecasted solar yield is now used to determine the charging duration.

Unifies the calculation of charging efficiency

## Approved roadmap item

Discussion / issue: #

## Tests

Tested on my setup at home


ffunes:
Thanks for digging into this — the diagnosis in the first commit is real: `charge_time_h = energy_needed / (P_max × CHARGE_EFFICIENCY)` assumes rated power, which is nothing like the pace of solar-following PD. But I can't merge this as it stands, and I'd like the two commits split.

I checked both commits in a worktree against the test suite (64 tests green on your branch).

## 1. The branch needs a rebase

The PR is opened from your fork's `main`, which is 161 commits behind `release/v1.5.0`. The diff therefore shows 47 files / 364 deletions and would revert v1.5.0 work (including `manifest.json` back to `1.4.0`). Please rebase onto `release/v1.5.0` — the diff should touch 2 files.

## 2. "Unify the calculation of charging efficiency" — please drop this commit

`CHARGE_EFFICIENCY` is not an AC-to-DC energy ratio. `const/integration_const.py:558-564` documents it as a *power* derate used to size a charging window, and explicitly says it is distinct from the AC-to-AC energy ratio — which already exists as `DEFAULT_ROUND_TRIP_EFFICIENCY`, and is user-tunable.

`DELAY_SAFETY_FACTOR = 1.3` is the empirical cushion that already absorbs conversion loss. Dividing by 0.85 on top makes the effective factor **1.53**, and that is a behaviour change for every user, not a refactor. Measured on a sunny day (forecast 12 kWh, 5 kWh needed, net 7.42 kWh, default 1 h margin):

| hour | before | with this commit |
|---|---|---|
| 07:00 | delay held (est. 10:06) | **unlocks** (`energy_balance`) |

The delay breaks at dawn on a day the sun would have covered easily. The tell is already in the PR: four existing cushion tests had to be retuned (`_cushion_mgr(4.85)` → `5.5`).

The new test doesn't cover the change either. With the AC conversion disabled, `test_energy_balance_gate_accounts_for_charge_efficiency` still passes (only the 3 retuned tests fail) — with `forecast = needed = 3.0` the pre-change code already fell into the cushion branch and unlocked with `energy_balance`.

If the forecast looks too optimistic in practice, that's a tuning discussion on `DELAY_SAFETY_FACTOR`, not a second derate stacked on the first.

## 3. The time-backup commit is a no-op at the default safety margin

I replaced `min(nominal_time_backup_unlock_h, solar_feasible_unlock_h)` with `nominal_time_backup_unlock_h` and **all 64 tests still pass**, including the new ones. An hour-by-hour sweep with the default 1 h margin produces output identical to the baseline.

The reason: `energy_insufficient` (net over `[now, t_end]` vs `needed × 1.3`) is evaluated first and always fires before the new bare `× 1.0` test over `[now, deadline]`, because the last safety-margin hours of the cosine curve deliver far less than `0.3 × needed`. It only bites at large margins (with 4 h: unlock at 11:00 instead of 13:00).

So the improvement you saw at home most likely came from the efficiency commit, by accident. `test_solar_feasible_unlock_pulls_time_backup_earlier_than_full_power_assumption` asserts `unlock_h <= nominal`, which was already true before the change, since `est_unlock_h = min(time_backup_unlock_h, energy_balance_unlock_h)`.

Two more things on that commit:

- **Missing `<= now_h` guard.** `energy_balance_unlock_h` is discarded when it lands at or before `now_h` while the measured balance disagrees (`charge_delay.py:781-786`). `solar_feasible_unlock_h` has no such guard and feeds `time_limit_reached = now_h >= time_backup_unlock_h` directly. `_estimate_energy_balance_unlock_h` returns `now_h` as soon as `net_solar_at(now_h) < threshold`, and the sinusoid can disagree with the measured `remaining_solar_kwh` (`forecast − produced_today`), so this is a path to an immediate same-day unlock with reason `time_backup`.
- **Edge case:** `horizon = max(t_start, min(horizon_h, t_end))` collapses to `t_start` when `t_end − margin <= t_start` (short winter day, large margin) — net goes negative and the delay unlocks on the spot.

Also, `status["energy_needed_kwh"]` stays battery-side while every comparison moves to AC, so the sensor attributes and `net_solar_kwh` no longer explain the decision shown on the dashboard.

## What I'd need to merge

1. Rebase onto `release/v1.5.0`.
2. Drop the efficiency commit.
3. Keep the time-backup commit with: a test that **fails** without the `min()`, the `<= now_h` guard, and a real-world trace of the case that motivated the PR — because at the default margin this change currently does nothing.

Minor: the new comment blocks run ~20 lines for 6 lines of code, well above the density of the rest of the file. Happy to keep the reasoning, just shorter.

PR #455, opened 2026-09-11
fix(pricing): credit the whole day's sun to the discharge reserve (#446)

### LINKED ISSUE #446: Discharge reserve holds the battery overnight on a day with 13.9 kWh of forecast PV
Last night the discharge reserve parked my battery at 48% from 03:37 to 05:45 while the house ran on the grid. The solar forecast for the day was 13.9 kWh, so there was never any doubt the reserve could be refilled. At 05:45 it released with `pv_refills_in_time` and discharged the same energy at the same price it had just refused to use.

So the hold cost me about 0.5 kWh of import and gained nothing.

## Setup

Omnibattery v1.5.0b1, one Marstek Venus E (5.12 kWh, single phase), dynamic pricing via Zonneplan (NL, 15-minute slots). Discharge Price Floor 0.34 €/kWh, reserve min saving 0.05, discharge cutoff 12%, Charge Delay on.

## What happened (2026-09-11, Europe/Amsterdam)

| time | state | reason | reserve | usable | free space | price |
|---|---|---|---|---|---|---|
| 00:00:09 | on | `reserved_for_dearer_hours` | 2.084 | 2.714 | 1.536 | 0.35337 |
| 00:10:09 | off | `pv_refills_in_time` | 0 | 2.714 | 1.536 | 0.35337 |
| 00:45:09 | on | `reserved_for_dearer_hours` | 1.485 | 2.560 | 1.690 | 0.34953 |
| 01:00:09 | off | `pv_refills_in_time` | 0 | 2.458 | 1.792 | 0.36900 |
| 01:15:10 | on | `reserved_for_dearer_hours` | 1.543 | 2.406 | 1.843 | 0.34629 |
| 05:45:11 | off | `pv_refills_in_time` | 0 | 1.843 | 2.406 | 0.34404 |

SOC went 65% at 00:00 down to 48% at 03:25, then sat flat at 48% until 05:49. PD control quality read `blocked` from 03:45 to 05:45. Once the reserve released, SOC fell from 48% to 34% between 05:49 and 08:17.

Prices were flat across all of it. About 0.345 €/kWh during the hold, 0.34 to 0.36 during the discharge right after. The evening peak it was saving for is 0.52 at 19:45.

## The 00:00 and 00:10 pair

Those two are ten minutes apart and agree on every input the sensor exposes: usable 2.714, free space 1.536, price 0.35337, same price slot, same horizon. One reserves 2.084 kWh, the other releases because PV will refill it. Nothing in the attributes explains the difference, which also makes it impossible to explain to a user staring at the entity.

## Where I think it comes from

In `pricing/discharge_reserve.py` the PV credit is:

```python
recharge = self._expected_recharge_kwh(now, claiming[0].start, free_space_kwh)
reserve = max(0.0, reserve - recharge)
```

```python
surplus += _positive(slot.expected_surplus_kwh)
return min(space, surplus)
```

Two things about that bother me.

The window stops at `claiming[0].start`, the *earliest* claiming slot. One pre-dawn slot clearing `current_price + min_saving` is enough to close the window before sunrise, and then the credit is zero no matter how much sun the day holds. Since the threshold is rebuilt from the current price every cycle, a small price wobble can drag that first slot across dawn and back. That would explain the flapping.

And the credit is capped at the free space you have *now*. Overnight that number is small exactly when the reserve gets evaluated, and it grows as the battery drains. A fuller battery gets less PV credit than an emptier one, which is backwards.

## Questions

- Should the credit run to the last claiming slot instead of the first, or be an energy balance over the whole horizon?
- Should the free space cap use the space expected at PV time rather than right now?
- Is some hysteresis wanted so the reserve stops changing its mind while the price walks around the threshold?

## Logs

I only turned on debug for `control.discharge_reserve` after the fact, so I have no per-cycle trace for last night. It runs now and I will attach a full night as soon as this repeats. Raising it early so it is on your radar. Happy to do the PR if you tell me which direction you want.

syphernl:
Fixes #446. Steps 1 and 2.

The PV credit window ended at `claiming[0].start`, so one slot clearing
`current_price + min_saving` before sunrise closed it before any sun arrived and the
credit came out zero. On 2026-09-11 that held a 5.12 kWh Venus E at 48% from 03:37 to
05:45 on a 13.9 kWh forecast.

Claims are now walked in start order against a running pool of expected surplus. Each
draws from the sun landing before it, and no kWh pays twice. Room for that sun is the room
at PV time, interim demand included, clamped to `free_space + usable`.

The walk is per slot, not per interval. Demand after a sunny slot cannot make room for
that slot's surplus, which the interval version I wrote first got wrong.

Tests: four new in `tests/test_discharge_reserve.py`, all failing on `release/v1.5.0`.
Your acceptance case drops the reserve from 2.5 to 0.5 kWh at 00:30. The remaining 0.5 is
the morning claim, which no sun precedes, so it is not a full release. `pytest -q` gives
2475 passed and 10 skipped. `test_pruning_a_future_dated_day_rebuilds_the_aggregate` fails
on a clean `release/v1.5.0` too.

Two I left alone. `room += net_demand` counts slots that `price_discharge` will not let
the battery serve, so a blocked night opens room that never appears. Counting only slots
above the candidate threshold fixes it, but the planner does not know the discharge floor.
And `surplus_price_hold` can defer absorption past the claim it was credited to. Both
yours to call.

Step 4 separately. Step 3 after a few nights.

Approved roadmap item: n/a, bug fix. Direction agreed in #446.


ffunes (changes requested):
The chronological sweep is right. Earliest-claim-first against a FIFO pool is the correct greedy for "sun only pays the hours after it", and the new credit is always >= the old one, so this can only ever lower a reserve — the safe direction for a regression. 42/42 in `test_discharge_reserve.py`, and the full suite is 2498 passed / 10 skipped once the branch is current.

Three things.

**Rebase.** The branch sits on `c22c857`, not on the head of `release/v1.5.0`. `test_pruning_a_future_dated_day_rebuilds_the_aggregate` is already fixed on base by `dadfee9`, so that note is stale rather than wrong. Merging base in only conflicts on `CHANGELOG.md`.

**The trailing loop never runs.** Every claim comes out of `self.slots`, which `build_reserve_slots` sorts, so the inner `while` has settled all of them by the last slot. An `assert pending == len(claims)` in that position survives all 106 reserve tests. Drop lines 197-199.

**The credit ceiling wants a haircut.** Going from `free_space` now to `room + available` also removed the accidental cap on trusting the forecast. A 3 kWh evening claim at 00:30 against a forecast 8 kWh of midday surplus with 1 kWh of room: `release/v1.5.0` reserves 2.0 kWh, this reserves 0.0 and returns `pv_refills_in_time`. Through the day the 5-minute rebuild walks that back as the forecast firms up, but overnight there is nothing to correct it — a day-ahead figure releases the whole band, and a cloudy morning buys the peak at peak price. Please scale `expected_surplus_kwh` by ~0.75 in `_demand_and_surplus_by_slot` before it reaches the planner. The planner itself is fine as written.

Your two open ones I agree with, and neither blocks. `room += net_demand` over slots `price_discharge` will not serve is real; the discharge threshold could be handed to the planner the way `min_saving` already is, but that is its own change. `surplus_price_hold` deferring absorption past the claim it paid for is worth a separate issue.

One smaller note, pre-existing but load-bearing now: `free_space_kwh(fleet)` counts batteries with `can_discharge=False` while `usable_energy_kwh(fleet)` does not, so `ceiling` can exceed the physical band of the fleet that actually discharges. It only mattered as a display figure before; it is the credit cap now.

syphernl:
All three done, thanks.

Rebased onto the head of `release/v1.5.0`; only `CHANGELOG.md` conflicted. The
pruning test passes on the merged base, so that note of mine was stale.

Dropped the trailing loop. Your reasoning holds: every claim is one of the slots
the outer walk visits, so the inner `while` has drained `pending` by the last
pass. I checked it against the old version over 20k randomised cases, no
mismatches.

`expected_surplus_kwh` now reaches the planner at 75%, as a named constant in
`_demand_and_surplus_by_slot`. The planner is untouched. One thing I want to
flag: the haircut is flat, so it also applies at 14:00 on a remaining-today
sensor with almost no uncertainty left, and there it holds back a kWh the sun
has room for. Tapering on hours-to-first-PV, or keying off
`_forecast_is_remaining`, would aim it at the overnight case you describe.
Happy either way, say which and I will push it.

`free_space_kwh` counting `can_discharge=False` batteries while
`usable_energy_kwh` does not: agreed, and it is a real cap now rather than a
display figure. Separate PR, unless you would rather have it here.

I will open an issue for `surplus_price_hold` deferring absorption past its
claim.

Tests: `pytest -q` gives 2506 passed, 10 skipped.

ffunes (approved):
All three land. Rebase is on `7141ed0`, the loop removal is sound — `build_reserve_slots` sorts by `start` and every claim comes out of `self.slots`, so the inner `while` has drained `pending` by the last pass — and the haircut sits on `expected_surplus_kwh` alone, with `net_demand_by_slot` still on the full solar. Planner untouched. `pytest -q` gives 2506 passed and 10 skipped here too.

**Keep the haircut flat.** Two reasons.

`_forecast_is_remaining` does not measure uncertainty. A remaining-today sensor at 00:30 is still a day-ahead figure with the whole day in front of it, so keying off it aims at the wrong thing.

Tapering on hours-to-first-PV is the right idea in principle, but it opens a second tuning surface to correct an error whose worst case is over-reserving an afternoon — you lose the spread of one cheap hour. The no-haircut case loses the peak at peak price. The asymmetry is clear and 0.75 flat is already on the right side of it. If reports of a battery under-discharging in the afternoon show up, we revisit with data rather than with a curve.

`free_space_kwh` counting `can_discharge=False` batteries: separate PR, yes.

PR #488, opened 2026-09-19
fix(charge delay): publish energy_needed_kwh while waiting for solar

syphernl:
## What

While the charge delay is holding for sunrise, `sensor.*_charge_delay_status` dropped
`energy_needed_kwh`: the "No T_start detected" branch returns before the energy balance
is calculated, and the sensor omits `None` attributes. A dashboard that defaults the
missing attribute to 0 then reports "0.0 kWh needed to reach 95%" at an SOC nowhere near
the target, which reads as "nothing to do" instead of "not calculated yet".

The deficit is plain SOC arithmetic and does not depend on `T_start`, so it is now
published in that branch too, clamped at 0 so a battery already above target reports no
deficit rather than a negative one. The expression moved into `_energy_needed_kwh()` and
is reused at the existing call site. The delay decision itself is unchanged.

## Approved roadmap item

This is a bugfix.

## Tests

`uv run --with-requirements requirements-test.txt pytest tests/` — 1954 passed, 10 skipped.

Two new cases in `tests/test_charge_delay.py`:
- `test_waiting_for_solar_publishes_energy_needed` — fails on `main` (attribute absent),
  passes with the change.
- `test_waiting_for_solar_energy_needed_never_negative` — battery above target reports `0.0`.


ffunes:
Verified locally: cherry-picked onto `release/v1.5.0` (clean), both new tests fail without the change (`KeyError: 'energy_needed_kwh'`) and pass with it, full suite 2565 passed / 10 skipped.

Right root cause, minimal diff. Two things before merge:

1. **The same gap is still open in the other pre-sunrise hold.** `_low_forecast_price_release` holds at `charge_delay.py:1026` with state `Delayed (cheap import HH:MM est.)` and no `energy_needed_kwh` — and it runs *before* the branch this PR fixes (called at `:646`). A dashboard hits the identical "0.0 kWh needed" there. Worth publishing the deficit in that hold too, same call. (The `_price_delay` hold at `:895` is already covered: it runs after `:753`.)

2. **CHANGELOG entry missing** — one bullet under `[Unreleased]` / `Fixed`, with the usual "Thanks to @syphernl".

3. **Please retarget to `release/v1.5.0`** — that is where the unreleased fixes are landing, and the CHANGELOG bullet belongs there too. Rebase the branch onto it rather than merging: your commit alone applies cleanly (I verified with a cherry-pick), whereas merging the current branch as-is drags in the 17 commits `main` has ahead of the fork point and conflicts on unrelated files.

Minor note, not blocking: `energy_needed_kwh` is listed in `_DECISION_STATUS_FIELDS`, i.e. a completed-decision field that gets cleared in the setpoint phase. Publishing it from an incomplete hold bends that convention, but it is harmless in practice — `_solar_t_start` resets daily, so during "Waiting for solar" the other decision fields are still `None` and no stale figures mix in.

syphernl:
All three addressed.

1. The deficit is published once, before the `_charge_delay_balance_needs_charge` branch,
   so the cheap-import hold carries it too. The raw value still drives `batteries_full`.
2. CHANGELOG entry added under `[Unreleased]` / `Fixed`.
3. The convention note is not only cosmetic: `_charge_delay_balance_needs_charge` can flip
   mid-day, and the cheap-import hold then returns with a fresh deficit beside the previous
   cycle's `net_solar_kwh`. The same point now clears `_BALANCE_STATUS_FIELDS`, each of which
   is rewritten as soon as it is computed.

Tests: `pytest tests/` — 1956 passed, 10 skipped. Three new cases, each failing on `main`.

